PR #1, opened 2019-03-12
Removed numpy and matplotlib from packages

davclark:
I'm not sure why those were in there... but it breaks the installation, because there are not additional packages in directories called `matplotlib` and `numpy`.

Seems to build and install OK now (though it's not clear where I'm supposed to get sample data... will open a separate issue for that): https://gigantum.com/tinydav/nwb-ipy-widget-testing


bendichter:
oh thanks that was a mistake I meant to put them in install_requires

PR #10, opened 2019-05-16
vis spec no longer global

### LINKED ISSUE #7: custom neurodata_vis_spec should be preserved
several subwidgets call back to nwb2widget, but don't pass on the vis spec. For instance:
https://github.com/NeurodataWithoutBorders/nwb-jupyter-widgets/blob/49a6e64c1beaec9fdb6e39209e0d21bf77606be1/nwbwidgets/base.py#L44

This means that (most of) a custom spec just gets ignored.

I'm currently working on this issue if there are no objections.

NileGraddis:
fix #7 

allow vis specs passed to nwb2widget to be preserved through the recursion.

### DIFF
--- a/examples/YutaMouse41.ipynb
+++ b/examples/YutaMouse41.ipynb
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "code",
-   "execution_count": 3,
+   "execution_count": 1,
    "metadata": {},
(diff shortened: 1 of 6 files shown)


bendichter:
You don't have to make a deep copy to make changes to the vis_spec, right?

PR #41, opened 2020-01-28
Modification in plot_traces()

Armin12:


nwbwidgets/base.py
```diff
@@ -213,51 +212,52 @@ def plot_traces(time_series: TimeSeries, time_start=0, time_duration=None, trace
...
+    
     if time_duration is None:
         t_ind_stop = None
     else:
         t_ind_stop = timeseries_time_to_ind(time_series, time_start + time_duration)
-
-    if trace_window is None:
-        trace_window = [0, time_series.data.shape[1]]
-    tt = get_timeseries_tt(time_series, t_ind_start, t_ind_stop)
-    if time_series.data.shape[1] == len(tt):  # fix of orientation is incorrect
```
bendichter: This line is meant to check if the data was accidentally stored with time in the second dimension and switch the data if that is the case.

nwbwidgets/base.py
```diff
@@ -213,51 +212,52 @@ def plot_traces(time_series: TimeSeries, time_start=0, time_duration=None, trace
...
+    mini_data = time_series.data[:, t_ind_start:t_ind_stop]
+    
     gap = np.median(np.nanstd(mini_data, axis=0)) * 20
-    offsets = np.arange(trace_window[1] - trace_window[0]) * gap
-
+    offsets = np.arange(t_ind_stop - t_ind_start) * gap
     fig, ax = plt.subplots()
     ax.figure.set_size_inches(12, 6)
-    ax.plot(tt, mini_data + offsets)
+    ax.plot(tt, (mini_data + offsets).T)
```
bendichter: I think this was right before. TimeSeries.data objects are always time first: https://github.com/NeurodataWithoutBorders/nwb-schema/blob/5d6160eccc6e4c786fc328c8f1d227420f933bda/core/nwb.base.yaml#L69-L79

nwbwidgets/base.py
```diff
@@ -213,51 +212,52 @@ def plot_traces(time_series: TimeSeries, time_start=0, time_duration=None, trace
...
+    
+    if t_ind_stop == None:
+        tt = get_timeseries_tt(time_series, t_ind_start, time_series.data.shape[-1])
     else:
-        mini_data = time_series.data[t_ind_start:t_ind_stop, trace_window[0]:trace_window[1]]
-
+        tt = get_timeseries_tt(time_series, t_ind_start, t_ind_stop)
+    
+    if trace_window is None:
+        trace_window = [0, time_series.data.shape[-1]]
```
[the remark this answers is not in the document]
Armin12: Done!

nwbwidgets/base.py
Armin12: Correct.

nwbwidgets/base.py
```diff
@@ -213,51 +212,52 @@ def plot_traces(time_series: TimeSeries, time_start=0, time_duration=None, trace
...
         t_ind_stop = timeseries_time_to_ind(time_series, time_start + time_duration)
-
-    if trace_window is None:
-        trace_window = [0, time_series.data.shape[1]]
-    tt = get_timeseries_tt(time_series, t_ind_start, t_ind_stop)
-    if time_series.data.shape[1] == len(tt):  # fix of orientation is incorrect
-        mini_data = time_series.data[trace_window[0]:trace_window[1], t_ind_start:t_ind_stop].T
+    
+    if t_ind_stop == None:
+        tt = get_timeseries_tt(time_series, t_ind_start, time_series.data.shape[-1])
```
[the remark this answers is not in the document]
Armin12: Using time in the first dimension of the TimeSeries.data object solved the problem.

nwbwidgets/base.py
Armin12: Correct.

nwbwidgets/test/test_timeseries.py
```diff
@@ -11,3 +11,32 @@ def test_timeseries_widget():
...
+    
+import numpy as np
+from pynwb import TimeSeries
+from nwbwidgets.base import show_text_fields
+
+
+def test_show_text_fields():
+    
+    data = np.array([[1., 2., 3., 4., 1., 2., 3., 4.],
+                               [11., 12., 13., 14., 11., 12., 13., 14.]])
```
bendichter: This array has the wrong orientation

nwbwidgets/base.py
```diff
@@ -213,12 +212,15 @@ def plot_traces(time_series: TimeSeries, time_start=0, time_duration=None, trace
...
     -------
-
     """
-
+    
+    if type(time_series.data) != np.ndarray:
+        time_series.data = np.asarray(time_series.data)
+    if len(time_series.data.shape) == 1:
+        time_series.data = time_series.data.reshape((-1,1))
+
```
bendichter: Good ideas. However, both of these lines are going to be problematic, because they both require reading the data. When you write an NWB file and then read it back, it presents datasets not as lists or np.arrays, but has `h5py.Dataset`s. The data is not read, but the `h5py.Dataset`s provide "views" into the data. The data is only read when you try to slice it (e.g. `time_series.data[:5]`). This is important because the data is sometimes too big to fit into memory, so we only want to read data that we are planning to view. 

Also, I would rather not change an attribute of `time_series.data`. In python, that change will persist in that object, even outside the scope of the function. Instead, we should declare a variable that is local to the function, e.g. `data = time_series.data` and use that.

bendichter: Maybe a better approach is to declare a `n_traces` variable instead of using `time_series.data.shape[1]`.

bendichter: as you might assume from the name of the function (plot_traces), I wasn't really planning on visualizing a single trace, but yeah I think it is better if this function does not break in that case.

nwbwidgets/base.py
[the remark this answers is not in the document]
bendichter: I think it would be better to define `n_clusters` and use that. This `reshape` call is inefficient because it may require reading all the data.

nwbwidgets/test/test_timeseries.py
```diff
@@ -11,3 +11,31 @@ def test_timeseries_widget():
...
+from pynwb import TimeSeries
+from nwbwidgets.utils.timeseries import get_timeseries_tt
+
+
+def test_get_timeseries_tt():
+    
+    data = list(range(100, 200, 10))
+    ts = TimeSeries(name='test_timeseries', data=data, unit='m', starting_time=0.0, rate=1.0)
+    
+    get_timeseries_tt(ts)
```
bendichter: For functions that have a numeric output, test that the output is correct with `np.testing.assert_array_equal`

nwbwidgets/base.py
bendichter: @Armin12  please address this

nwbwidgets/test/test_timeseries.py
bendichter: @Armin12  please address this

PR #45, opened 2020-02-03
Unit test and plot_traces changes

Armin12:


nwbwidgets/base.py
```diff
@@ -198,10 +198,11 @@ def show_text_fields(node, exclude=('comments', 'interval'), **kwargs):
     return widgets.VBox(info)
 
 
-def plot_traces(time_series: TimeSeries, time_start=0, time_duration=None, trace_window=None,
-                title: str = None, ylabel: str = 'traces'):
+def plot_traces(time_series: TimeSeries, time_start = 0, time_duration = None, trace_window = None,
+                n_trace = 1, title: str = None, ylabel: str = 'traces'):
```
bendichter: 1) don't put spaces around `=` in function calls
2) n_traces should be computed from trace_window, not supplied in the function call

nwbwidgets/test/test_timeseries.py
```diff
@@ -11,3 +11,127 @@ def test_timeseries_widget():
...
+from pynwb import NWBFile
+from nwbwidgets.base import processing_module
+from datetime import datetime
+from dateutil.tz import tzlocal
+from pynwb.behavior import Position
+from pynwb.behavior import SpatialSeries
+from pynwb import ProcessingModule
+
+
+def test_processing_module():
```
bendichter: test_processing_module should not be included in this PR because it has nothing to do with timeseries changes and tests

Armin12: What about fig2widget() and vis2widget()? I think they should be moved to another file with another PR.

nwbwidgets/base.py
Armin12: For (2), when the `trace_window` is `None` , it needs to calculate `time_series.data.shape[1]`. We can make the `trace_window` mandatory in the function.

nwbwidgets/test/test_timeseries.py
bendichter: yes, please separate changes into distinct PRs where appropriate

nwbwidgets/test/test_timeseries.py
```diff
@@ -11,3 +11,82 @@ def test_timeseries_widget():
...
+
+
+def test_fig2widget():
+    
+    data = np.random.rand(160,3)
+    
+    fig=plt.figure(figsize=(10, 5))
+    plt.plot(data)
+
+    fig2widget(fig)
```
bendichter: confirm that type is widget

bendichter: also this should be in a different PR

nwbwidgets/base.py
```diff
@@ -199,9 +199,10 @@ def show_text_fields(node, exclude=('comments', 'interval'), **kwargs):
 
 
 def plot_traces(time_series: TimeSeries, time_start=0, time_duration=None, trace_window=None,
-                title: str = None, ylabel: str = 'traces'):
+                n_trace=1, title: str = None, ylabel: str = 'traces'):
```
bendichter: please remove n_traces

nwbwidgets/test/test_timeseries.py
```diff
@@ -11,3 +11,79 @@ def test_timeseries_widget():
...
+    show_timeseries(ts, istart=5, istop=56)
+
+    
+    
+import numpy as np
+from pynwb import TimeSeries
+from nwbwidgets.base import show_subject
+
+
+def test_show_subject():
```
bendichter: show_subject is not relevant and should be in a different PR

nwbwidgets/base.py
bendichter: you can use `len(time_series.data[0])` if the `shape` attribute is not available.

Armin12: Yes, but still calculating `mini_data` needs multi-axis slicing which may require for loop or other slicing operators when `time_series.data` is not a numpy array.

nwbwidgets/base.py
```diff
@@ -213,12 +215,12 @@ def plot_traces(time_series: TimeSeries, time_start=0, time_duration=None, trace
...
     title: str, optional
     ylabel: str, optional
-
     Returns
     -------
-
     """
-
+    data = time_series.data
+    data = np.asarray(data)
```
bendichter: You've already tried this. It is still not correct!

bendichter:
The changes to nwbwidgets/test/test_timeseries.py  are right. Please remove all other changes.

bendichter:
The time series plotting code was moved to its own file: time_series.py. Here, when you merged your changes, you kept the functions in the old place, which is duplicating that code now in two places. Please remove all the plotting code, and just keep the tests.

PR #46, opened 2020-02-03
Test Processing Module

Armin12:


nwbwidgets/test/test_processing_module_widgets.py
```diff
@@ -0,0 +1,84 @@
...
+    neurodata_vis_spec =	{'ndtype': 'str'}
+    
+    # Test processing_module function in base.py
+    processing_module(nwbfile.processing['behavior'], neurodata_vis_spec)
+
+
+    
+    
+
+import numpy as np
```
bendichter: imports should go at the top of the file

nwbwidgets/test/test_processing_module_widgets.py
```diff
@@ -0,0 +1,84 @@
...
+from pynwb.behavior import Position
+from pynwb.behavior import SpatialSeries
+from pynwb import ProcessingModule
+
+
+def test_processing_module():
+    
+    # Create a NWB file
+    start_time = datetime(2020, 1, 29, 11, tzinfo=tzlocal())
+    nwbfile = NWBFile(session_description = 'Test Session',
```
bendichter: I use PEP8 as a style guide, which advises ["Don't use spaces around the = sign when used to indicate a keyword argument, or when used to indicate a default value for an unannotated function parameter."](https://www.python.org/dev/peps/pep-0008/#other-recommendations)

nwbwidgets/test/test_processing_module_widgets.py
```diff
@@ -0,0 +1,84 @@
+
+import numpy as np
+from pynwb import NWBFile
+from nwbwidgets.base import processing_module
+from datetime import datetime
+from dateutil.tz import tzlocal
+from pynwb.behavior import Position
```
bendichter: the Position and SpatialSeries imports should be on the same line

PR #47, opened 2020-02-03
Test graphical functions

Armin12:


nwbwidgets/test/test_plot_func.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def test_fig2widget():
+    
+    data = np.random.rand(160,3)
+    
+    fig=plt.figure(figsize=(10, 5))
+    plt.plot(data)
+
+    fig2widget(fig)
```
bendichter:
```suggestion
```

There's no reason to call this twice.

nwbwidgets/test/test_plot_func.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+
+import numpy as np
+import matplotlib.pyplot as plt
+from ipywidgets import widgets
+from nwbwidgets.base import vis2widget
+
+
+def test_vis2widget():
```
bendichter: vis2widget can take a widget or a figure. You should have a test for each. This one should be called something like `test_vis2widget_input_widget`

nwbwidgets/test/test_plot_func.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+def test_fig2widget():
+    
+    data = np.random.rand(160,3)
+    
+    fig=plt.figure(figsize=(10, 5))
+    plt.plot(data)
+    
+    isinstance(fig2widget(fig), widgets.Widget)
```
bendichter: this will simply return True or False. You want to `assert` that this is true

nwbwidgets/test/test_plot_func.py
```diff
@@ -0,0 +1,155 @@
...
+    max=10,
+    step=1,
+    description='Test:',
+    disabled=False,
+    continuous_update=False,
+    orientation='horizontal',
+    readout=True,
+    readout_format='d')
+
+    vis2widget(wg)
```
bendichter: use `assert` here as well

nwbwidgets/test/test_plot_func.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+def test_vis2widget_input_figure():
+    
+    data = np.random.rand(160,3)
+    
+    fig=plt.figure(figsize=(10, 5))
+    plt.plot(data)
+    
+    vis2widget(fig)
```
bendichter: how is this different from `test_fig2widget`?

nwbwidgets/test/test_plot_func.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+def test_show_subject():
+    
+    data = np.random.rand(160,3)
+    ts = TimeSeries(name='test_timeseries', data=data, unit='m', starting_time=0.0, rate=1.0)
+
+    
+    show_subject(ts)
```
bendichter: you are running `show_subject` on a `TimeSeries` object. `show_subject` is meant to be run on a `Subject` object.

nwbwidgets/test/test_plot_func.py
```diff
@@ -0,0 +1,155 @@
...
+    
+    
+def test_show_dynamic_table():
+    
+    d = {'col1': [1, 2], 'col2': [3, 4]}
+    DT = DynamicTable.from_dataframe(df=pd.DataFrame(data=d), 
+                                     name='Test Dtable', 
+                                     table_description='no description')
+    
+    show_dynamic_table(DT)
```
bendichter: This test is good

nwbwidgets/test/test_plot_func.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+def test_dict2accordion():
+    
+    data = list(range(100, 200, 10))
+    ts = TimeSeries(name='test_timeseries', data=data, unit='m', starting_time=0.0, rate=1.0)
+    
+    beh_events = BehavioralEvents(time_series=ts)
+    
+    dict2accordion(beh_events.time_series, default_neurodata_vis_spec)
```
bendichter: `dict2accordion` takes a `dict` as an input

nwbwidgets/test/test_plot_func.py
```diff
@@ -0,0 +1,155 @@
...
+    # Create a data interface object (position)
+    spatial_series = SpatialSeries(name = 'position',
+                               data = np.linspace(0, 1, 20),
+                               rate = 50.,
+                               reference_frame = 'starting gate')
+    position = Position(spatial_series = spatial_series)
+    
+    
+    # Test show position function in behavior.py
+    show_position(position, default_neurodata_vis_spec)
```
bendichter: This test is good

nwbwidgets/test/test_plot_func.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+
+def test_show_behavioral_events():
+
```
bendichter: This test is good, but here and in other places only use 0 or 1 blank line between function definition and function code, not 2

nwbwidgets/test/test_plot_func.py
```diff
@@ -0,0 +1,155 @@
...
+    
+    
+    # Create a data interface object (position)
+    spatial_series = SpatialSeries(name = 'position',
+                               data = np.linspace(0, 1, 20),
+                               rate = 50.,
+                               reference_frame = 'starting gate')
+    
+    
+    # Test show_spatial_series function in behavior.py
```
bendichter: You should be able to share code from this function and the one above

nwbwidgets/test/test_plot_func.py
```diff
@@ -0,0 +1,155 @@
...
+    show_spatial_series(spatial_series)
+    
+    
+    
+    
+def test_show_annotations():
+    
+    data = np.random.rand(160,12)
+    ts = TimeSeries(name='test_timeseries', data=data, unit='m', starting_time=0.0, rate=1.0)
```
bendichter: this object appears to be unused

nwbwidgets/test/test_plot_func.py
```diff
@@ -25,32 +26,37 @@ def test_fig2widget():
...
+
+        fig=plt.figure(figsize=(10, 5))
+        plt.plot(data)
+
+        assert isinstance(vis2widget(fig), widgets.Widget)
+        
+    def test_vis2widget_input_other(self):
+        data = np.random.rand(160,3)
+        with pytest.raises(ValueError, match="unsupported vis type"):
+            vis2widget(data)
```
bendichter: nice!

nwbwidgets/test/test_plot_func.py
Armin12: Do you mean to group them into one class?

bendichter: I think this is fine for now. Let's get this through, and then we can talk about improving the structure with with a `setup` method later

PR #49, opened 2020-02-05
Update ecephys.py

Armin12:


nwbwidgets/ecephys.py
```diff
@@ -15,10 +15,10 @@ def show_spectrogram(neurodata, channel=0, **kwargs):
     fig, ax = plt.subplots()
     f, t, Zxx = stft(neurodata.data[:, channel], neurodata.rate, nperseg=2*17)
     ax.imshow(np.log(np.abs(Zxx)), aspect='auto', extent=[0, max(t), 0, max(f)], origin='lower')
-    ax.set_ylim(0, 50)
```
bendichter: this number probably depends on the input data. Just changing it to another hard-coded value does not fix the underlying issue

PR #62, opened 2020-02-17
Update test_processing_module_widgets.py

Armin12:


nwbwidgets/test/test_processing_module_widgets.py
```diff
@@ -7,52 +7,39 @@
...
-    nwbfile.add_processing_module(behavior_module)
+        behavior_module = ProcessingModule(name='behavior',
+                                                       description='preprocessed behavioral data')
+        nwbfile.add_processing_module(behavior_module)
 
-    nwbfile.processing['behavior'].add(position)
+        nwbfile.processing['behavior'].add(self.position)
 
-    processing_module(nwbfile.processing['behavior'], default_neurodata_vis_spec)
+        self.processing_module(nwbfile.processing['behavior'], default_neurodata_vis_spec)
```
bendichter: `self.processing_module` is never defined.

nwbwidgets/test/test_processing_module_widgets.py
```diff
@@ -7,52 +7,39 @@
...
 
+    def test_nwb2widget(self):
 
-def test_nwb2widget():
+        self.nwb2widget(self.position, default_neurodata_vis_spec)
```
bendichter: `nwb2widget` is not a method or attribute of this class. Should not have `self`.

nwbwidgets/test/test_processing_module_widgets.py
```diff
@@ -7,52 +7,39 @@
...
-    
+    def test_show_neurodata_base(self):
 
-    spatial_series = SpatialSeries(name = 'position',
-                               data = np.linspace(0, 1, 20),
-                               rate = 50.,
-                               reference_frame = 'starting gate')
-    position = Position(spatial_series = spatial_series)
-    
+        self.show_neurodata_base(self.position, default_neurodata_vis_spec)
```
bendichter: `show_neurodata_base` is not a method or attribute of this class. Should not have `self`.

PR #68, opened 2020-02-18
Update misc.py

Armin12:


nwbwidgets/misc.py
```diff
@@ -287,10 +289,10 @@ def show_psth(units: pynwb.misc.Units, index=0, start_label='start_time', before
             labels, cvals = np.unique(coldata, return_inverse=True)
         elif np.all(np.isreal(data)):
             coltype = 'numeric'
-            cvals = coldata
+            labels, cvals = np.unique(coldata, return_inverse=True)
```
bendichter: `labels` is only ever used if `coltype == 'cat'`, so I don't think this is necessary. What was the intention behind adding it?

PR #78, opened 2020-02-25
Update test_controllers.py

Armin12:


nwbwidgets/test/test_controllers.py
```diff
@@ -12,16 +12,30 @@ class MoveSliderTestCase(unittest.TestCase):
...
+        value=7,
+        min=2,
+        max=10,
+        step=1,
+        description='Test:',
+        disabled=False,
+        continuous_update=False,
+        orientation='horizontal',
+        readout=True,
+        readout_format='d')
```
bendichter: I would expect these lines to be indented

PR #88, opened 2020-03-03
Create test_utils_dynamictable.py

Armin12:


bendichter:
@Armin12 this technically works, but does not really capture the intention of the function. What we are trying to do here is infer which of the columns of a table hold categorical information (e.g. cell types or good vs bad). If you are randomly generating numbers in a continuous range, those are continuous variables, not categorical variables.

nwbwidgets/test/test_utils_dynamictable.py
```diff
@@ -0,0 +1,21 @@
...
+    data2 = np.array([False, False,  True,  True, False,  True, False,  True, False])
+    
+    vd1 = VectorData('spikes','vector data for creating a DynamicTable',data=data1)
+    vd2 = VectorData('LFP','vector data for creating a DynamicTable',data=data2)
+    vd=[vd1,vd2]
+    
+    dynamic_table = DynamicTable(name='test table',description='This is a test table',
+                                 columns=vd,colnames=['spikes','LFP'])
+    
+    assert(infer_categorical_columns(dynamic_table)['spikes'].shape[0]==2)
```
bendichter: test equivalence of the output to be the dictionary you expect

PR #97, opened 2020-03-10
Test ophys

Armin12:


nwbwidgets/ophys.py
```diff
@@ -65,17 +65,16 @@ def show_plane_segmentation_3d(plane_seg: PlaneSegmentation):
 
     nrois = len(plane_seg)
 
-    dims = np.array([max(max(plane_seg['voxel_mask'][i][dim]) for i in range(nrois))
-                     for dim in ['x', 'y', 'z']]).astype('int') + 1
+    dims = np.array([max(max([plane_seg['voxel_mask'][i][j][dim] for j in range(len(plane_seg['voxel_mask'][i]))]) for i in range(nrois)) for dim in [0, 1, 2]]).astype('int') + 1
```
Armin12: Two changes were made: 
1) j iterates over all the tuples in the voxel mask. It is needed because the number of tuples in a voxel mask list can differ for each ROI.
2) ['x', 'y', 'z'] changed to [0,1,2] because the third dimension of a list could not be of the type str.

Please let me know if it needs to be changed.

bendichter:
@Armin12 can you please resolve this conflict?

nwbwidgets/ophys.py
bendichter: I don't understand

PR #102, opened 2020-03-22
Add psth evaluation function

d-sot:
* add psth evaluation
 * update compute_smoothed_firing_rate


d-sot:
The minor update to `compute_smoothed_firing_rate` is due to `binned_spikes[np.searchsorted(tt, spike_times)] += 1`, since it could potentially miss out where multiple spike times would return the same index in `tt`.

PR #105, opened 2020-05-08
Add tqdm to requirements.txt

brianhhu:
Add tqdm as a required dependency


bendichter:
@brianhu yeah, thanks for the PR. In this case, you would also need to add this dependency to the `setup.py` file.

PR #107, opened 2020-06-08
Update icephys accordions with AlignedDynamicTable and renamings

lvsltz:
There have been some major changes since February in `ndx-icephys-meta` and this PR syncs `nwb-jupyter-widges` with that extension:
- some tables have been renamed
- `IntracellularRecordingsTable` is now an `AlignedDynamicTable`

### DIFF
--- a/nwbwidgets/icephys.py
+++ b/nwbwidgets/icephys.py
@@ -2,20 +2,20 @@ from .base import lazy_show_over_data, GroupingWidget
 from .timeseries import show_timeseries_mpl
 from ipywidgets import widgets
 import matplotlib.pyplot as plt
-from ndx_icephys_meta.icephys import SweepSequences
+from ndx_icephys_meta.icephys import SequentialRecordingsTable
 from functools import partial
 import numpy as np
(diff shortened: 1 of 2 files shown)


bendichter:
@lvsltz thanks, these changes look great!

Do you know if these changes were associated with version numbers? I'm worried that after we make these changes someone might install an older version of ndx-icephys-meta and then have the opposite problem. The best way to avoid this is to explicitly enforce specific versions in the requirements or setup.py files. I'll take a look at the ndx-icephys-meta repo to see if we can refer to specific release versions.

bendichter:
It looks like the last (and only) release was Feb 3, and many of these changes were made March 3-4, so they are unreleased. In general, I only want to depend on released code (on pypi), so I'm going to wait to merge this until those changes are in an official pypi release. Also, I'm going to add a version constraint for when this is eventually merged.

lvsltz:
Indeed, it makes sense to wait. I overlooked the pypi release.

bendichter:
This is very helpful nevertheless! Thanks for helping me keep on top of this!

PR #126, opened 2020-10-28
Initial code draft

MRScheid:
Creating place field class

### DIFF
--- a/nwbwidgets/placefield.py
+++ b/nwbwidgets/placefield.py
@@ -1,0 +1,291 @@
+import numpy as np
+from scipy.ndimage import label
+from scipy.ndimage.filters import gaussian_filter, maximum_filter
+
+import pynwb
+from pynwb.misc import Units
+from .behavior import plotly_show_spatial_trace
+from .base import vis2widget
+from ipywidgets import widgets
+from .utils.units import get_spike_times
+
+## To-do
+# [] Create PlaceFieldWidget class
+    # [X] Refactor place field calculation code to deal with nwb data type
+        # [X] Incorporate place field fxns into class
+        # [X] Change all internal attributes references
+        # [X]Change all internal method references
+
+    # [X] Get pos
+    # [X] Get time
+    # [X] Get spikes
+    # [] Get trials / epochs
+
+    # [] Submit draft PR
+
+    # [] Modify plotly_show_spatial_trace to plot 2D heatmap representing place fields or create new figure function?
+        # [] Dropdown that controls which unit
+
+    # [] Work in buttons / dropdowns / sliders to modify following parameters in place field calculation:
+        # [] Different epochs
+        # [] Gaussian SD
+        # [] Speed threshold
+        # [] Minimum firing rate
+        # [] Place field thresh (% of local max)
+
+# Put widget rendering here
+class PlaceFieldWidget(widgets.HBox):
+
+    def __init__(self, unit: Units, node: pynwb.behavior.SpatialSeries):
+
+        super().__init__()
+
+        # Initialize receptive fields
+        self.receptive_fields = np.zeros(firing_rate.shape, dtype=int)
+        # Get pos
+        pos, unit = get_timeseries_in_units(node)
+        # Get time
+        pos_tt = get_timeseries_tt(node)
+        # Get spikes
+        spikes = get_spike_times(unit)
+        # Pixel width?
+        pixel_width =
+
+        # Put widget controls here:
+            # - Gaussian SD
+            # - Speed threshold
+            # - Minimum firing rate
+            # - Place field thresh (% of local max)
+
+        self.compute_2d_firing_rate(pos, pos_tt, spikes, pixel_width) # speed_thresh=0.03, gaussian_sd=0.0184,
+                                    # x_start=None, x_stop=None, y_start=None, y_stop=None)
+
+        self.compute_2d_place_fields() # min_firing_rate=1, thresh=0.2,
+                                                    # min_size=100):
+
+
+        return vis2widget(plotly_show_spatial_trace(self.receptive_fields))
+
+
+    # Put place field code here
+    def find_nearest(self, arr, tt):
+        """Used for picking out elements of a TimeSeries based on spike times
+
+        Parameters
+        ----------
+        arr
+        tt
+
+        Returns
+        -------
+
+        """
+        arr = arr[arr > tt[0]]
+        arr = arr[arr < tt[-1]]
+        return np.searchsorted(tt, arr)
+
+    def smooth(self, y, box_pts):
+        """Moving average
+
+        Parameters
+        ----------
+        y
+        box_pts
+
+        Returns
+        -------
+
+        """
+        box = np.ones(box_pts) / box_pts
+        return np.convolve(y, box, mode='same')
+
+    def compute_speed(self, pos, pos_tt, smooth_param=40):
+        """Compute boolean of whether the speed of the animal was above a threshold
+        for each time point
+
+        Parameters
+        ----------
+        pos: np.ndarray(dtype=float)
+            in meters
+        pos_tt: np.ndarray(dtype=float)
+            in seconds
+        smooth_param: float, optional
+
+        Returns
+        -------
+        running: np.ndarray(dtype=bool)
+
+        """
+        speed = np.hstack((0, np.sqrt(np.sum(np.diff(pos.T) ** 2, axis=0)) / np.diff(pos_tt)))
+        return self.smooth(speed, smooth_param)
+
+
+    def compute_2d_occupancy(self, pos, pos_tt, edges_x, edges_y, speed_thresh=0.03):
+        """Computes occupancy per bin in seconds
+
+        Parameters
+        ----------
+        pos: np.ndarray(dtype=float)
+            in meters
+        pos_tt: np.ndarray(dtype=float)
+            in seconds
+        edges_x: array-like
+            edges of histogram in meters
+        edges_y: array-like
+            edges of histogram in meters
+        speed_thresh: float, optional
+            in meters. Default = 3.0 cm/s
+
+        Returns
+        -------
+        occupancy: np.ndarray(dtype=float)
+            in seconds
+        running: np.ndarray(dtype=bool)
+
+        """
+
+        sampling_period = (np.max(pos_tt) - np.min(pos_tt)) / len(pos_tt)
+        is_running = self.compute_speed(pos, pos_tt) > speed_thresh
+        run_pos = pos[is_running, :]
+        occupancy = np.histogram2d(run_pos[:, 0],
+                                   run_pos[:, 1],
+                                   [edges_x, edges_y])[0] * sampling_period  # in seconds
+
+        return occupancy, is_running
+
+
+    def compute_2d_n_spikes(self, pos, pos_tt, spikes, edges_x, edges_y, speed_thresh=0.03):
+        """Returns speed-gated position during spikes
+
+        Parameters
+        ----------
+        pos: np.ndarray(dtype=float)
+            (time x 2) in meters
+        pos_tt: np.ndarray(dtype=float)
+            (time,) in seconds
+        spikes: np.ndarray(dtype=float)
+            in seconds
+        edges_x: np.ndarray(dtype=float)
+            edges of histogram in meters
+        edges_y: np.ndarray(dtype=float)
+            edges of histogram in meters
+        speed_thresh: float
+            in meters. Default = 3.0 cm/s
+
+        Returns
+        -------
+        """
+
+        is_running = self.compute_speed(pos, pos_tt) > speed_thresh
+
+        spike_pos_inds = self.find_nearest(spikes, pos_tt)
+        spike_pos_inds = spike_pos_inds[is_running[spike_pos_inds]]
+        pos_on_spikes = pos[spike_pos_inds, :]
+
+        n_spikes = np.histogram2d(pos_on_spikes[:, 0],
+                                  pos_on_spikes[:, 1],
+                                  [edges_x, edges_y])[0]
+
+        return n_spikes
+
+
+    def compute_2d_firing_rate(self, pos, pos_tt, spikes,
+                               pixel_width,
+                               speed_thresh=0.03,
+                               gaussian_sd=0.0184,
+                               x_start=None, x_stop=None,
+                               y_start=None, y_stop=None):
+        """Returns speed-gated occupancy and speed-gated and
+        Gaussian-filtered firing rate
+
+        Parameters
+        ----------
+        pos: np.ndarray(dtype=float)
+            (time x 2), in meters
+        pos_tt: np.ndarray(dtype=float)
+            (time,) in seconds
+        spikes: np.ndarray(dtype=float)
+            in seconds
+        pixel_width: float
+        speed_thresh: float, optional
+            in meters. Default = 3.0 cm/s
+        gaussian_sd: float, optional
+            in meters. Default = 1.84 cm
+        x_start: float, optional
+        x_stop: float, optional
+        y_start: float, optional
+        y_stop: float, optional
+
+
+        Returns
+        -------
+
+        occupancy: np.ndarray
+            in seconds
+        filtered_firing_rate: np.ndarray(shape=(cell, x, y), dtype=float)
+            in Hz
+
+        """
+        # pixel_width=0.0092,
+        # field_len = 0.46
+        # edges = np.arange(0, field_len + pixel_width, pixel_width)
+
+        x_start = np.min(pos[:, 0]) if x_start is None else x_start
+        x_stop = np.max(pos[:, 0]) if x_stop is None else x_stop
+
+        y_start = np.min(pos[:, 1]) if y_start is None else y_start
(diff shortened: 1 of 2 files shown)


bendichter:
@MRScheid looks good! Can you create something like these panels?

placefields

MRScheid:
> 
> 
> @MRScheid looks good! Can you create something like these panels?
> 
> placefields
> 
> Taken from [nelpy](https://github.com/nelpy/nelpy). We aren't using this library. Maybe we could, but might mean reinventing a lot at this point

Yes, I'll give it a shot

bendichter:
@MRScheid can you also make sure speed gating is used for the 1D rate maps?

nwbwidgets/placefield.py
```diff
@@ -426,25 +443,95 @@ def __init__(self, spatial_series: pynwb.behavior.SpatialSeries, **kwargs):
...
+        # fig = ax.plot(xx, filtered_firing_rate, '-')
+        # ax.set_xlabel('x ({})'.format(self.unit))
+        # ax.set_ylabel('firing rate (Hz)')
 
         return fig
+
+def plot_tuning_curves1D(ratemap, bin_pos, ax=None, normalize=False, pad=10, unit_labels=None, fill=True, color=None):
+    """
+    WARNING! This function is not complete, and hence 'private',
+    and may be moved somewhere else later on.
```
bendichter, line 448: Can you use numpy docstring here?

e.g. https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html

bendichter: Can you use a checkbox to control `normalize`?

nwbwidgets/placefield.py
```diff
@@ -0,0 +1,501 @@
+import numpy as np
+from scipy.ndimage import label
+from scipy.ndimage.filters import gaussian_filter, maximum_filter
+
+import matplotlib.pyplot as plt
+import nelpy.plotting as nlp
```
bendichter, line 6: putting this import here would require us to list `nelpy` as a dependency of this project. I try not to include dependencies unless we have to. What methods are we using form nelpy? Do we need them? If so, see SpikeInterface's way of dealing with these types of dependencies: https://github.com/SpikeInterface/spikeextractors/blob/5b9851c3ef8abd3720b2f97d30bc26bc0511f8b8/spikeextractors/extractors/neuroscopeextractors/neuroscopeextractors.py#L9-L14

nwbwidgets/placefield.py
```diff
@@ -0,0 +1,501 @@
...
+        in seconds
+    smooth_param: float, optional
+
+    Returns
+    -------
+    running: np.ndarray(dtype=bool)
+
+    """
+    speed = np.hstack((0, np.sqrt(np.sum(np.diff(pos.T) ** 2, axis=0)) / np.diff(pos_tt)))
+    return smooth(speed, smooth_param)
```
bendichter, line 102: These types of functions should go in https://github.com/NeurodataWithoutBorders/nwb-jupyter-widgets/tree/master/nwbwidgets/analysis

nwbwidgets/placefield.py
```diff
@@ -0,0 +1,501 @@
...
+
+
+def compute_2d_occupancy(pos, pos_tt, edges_x, edges_y, speed_thresh=0.03):
```
bendichter, line 105: I think we should create an analysis/place_fields.py file for all of this analysis. I generally like to separate analysis from visualization code

nwbwidgets/view.py
```diff
@@ -46,7 +46,9 @@ def show_dynamic_table(node, **kwargs) -> widgets.Widget:
     pynwb.behavior.Position: behavior.show_position,
     pynwb.behavior.SpatialSeries: OrderedDict({
         'over time': timeseries.SeparateTracesPlotlyWidget,
-        'trace': behavior.plotly_show_spatial_trace}),
+        'trace': behavior.plotly_show_spatial_trace,
+        'rate map': placefield.PlaceFieldWidget,
+        '1d rate map': placefield.PlaceField_1D_Widget}),
```
bendichter, line 51: Could we dynamically route this so that it uses 1D if the SpatialSeries is 1D?

nwbwidgets/placefield.py
bendichter: In this case it looks like it's just doing simple matplotlib tick manipulation, so I don't think it's worth adding nelpy as a dependency

nwbwidgets/view.py
bendichter: See https://github.com/NeurodataWithoutBorders/nwb-jupyter-widgets/blob/81b8b4f977dc7d40dc4b0f188d783b94f224a4fc/nwbwidgets/ophys.py#L245-L249

bendichter:
@MRScheid could you allow the widget to optionally take as input a TimeSeries of velocity? Sometimes that is provided separately.

MRScheid:
> 
> 
> @MRScheid could you allow the widget to optionally take as input a TimeSeries of velocity? Sometimes that is provided separately.

I'm not sure I fully understand what you have in mind--would it be a TimeSeries of velocity in place of the position in the SpatialSeries?  Would I then need to derive approximate position from the TimeSeries of velocity?

MRScheid:
Also, what would be the file extension of the TimeSeries input?  .nwb?  Maybe this will be good to discuss during our meeting today

bendichter:
Sometimes velocity is calculated separately with special smoothing function or something and I want to be able to use this in the widget if it is available.  The `TimeSeries` would be an additional optional input that would be supplied in addition to `SpatialSeries` representing position. It would not be part of an independent file. If it is not supplied, the function would calculate speed as it does now.

nwbwidgets/analysis/placefields.py
```diff
@@ -55,7 +55,7 @@ def compute_speed(pos, pos_tt, smooth_param=40):
     return smooth(speed, smooth_param)
 
 
-def compute_2d_occupancy(pos, pos_tt, edges_x, edges_y, speed_thresh=0.03):
+def compute_2d_occupancy(pos, pos_tt, edges_x, edges_y, speed_thresh=0.03, velocity=[]):
```
bendichter, line 58: An optional variable usually takes default value `None` and that's what should be used here.

nwbwidgets/analysis/placefields.py
```diff
@@ -89,7 +93,7 @@ def compute_2d_occupancy(pos, pos_tt, edges_x, edges_y, speed_thresh=0.03):
     return occupancy, is_running
 
 
-def compute_2d_n_spikes(pos, pos_tt, spikes, edges_x, edges_y, speed_thresh=0.03):
+def compute_2d_n_spikes(pos, pos_tt, spikes, edges_x, edges_y, speed_thresh=0.03, velocity=[]):
```
bendichter, line 96: An optional variable usually takes default value `None` and that's what should be used here.

nwbwidgets/analysis/placefields.py
```diff
@@ -129,7 +136,8 @@ def compute_2d_firing_rate(pos, pos_tt, spikes,
                            speed_thresh=0.03,
                            gaussian_sd=0.0184,
                            x_start=None, x_stop=None,
-                           y_start=None, y_stop=None):
+                           y_start=None, y_stop=None,
+                           velocity=[]):
```
bendichter, line 140: An optional variable usually takes default value `None` and that's what should be used here.

nwbwidgets/analysis/placefields.py
```diff
@@ -80,7 +80,11 @@ def compute_2d_occupancy(pos, pos_tt, edges_x, edges_y, speed_thresh=0.03):
     """
 
     sampling_period = (np.max(pos_tt) - np.min(pos_tt)) / len(pos_tt)
-    is_running = compute_speed(pos, pos_tt) > speed_thresh
+    if len(velocity) == 0:
+        is_running = compute_speed(pos, pos_tt) > speed_thresh
+    else:
+        is_running = self.velocity > speed_thresh
```
bendichter, line 86: you should compute speed from velocity (https://numpy.org/doc/stable/reference/generated/numpy.linalg.norm.html)

nwbwidgets/analysis/placefields.py
```diff
@@ -0,0 +1,263 @@
...
+    Returns
+    -------
+    occupancy: np.ndarray(dtype=float)
+        in seconds
+    running: np.ndarray(dtype=bool)
+
+    """
+
+    sampling_period = (np.max(pos_tt) - np.min(pos_tt)) / len(pos_tt)
+    if len(velocity) == 0:
```
bendichter, line 83:
```suggestion
    if velocity) is not None:
```

nwbwidgets/analysis/placefields.py
```diff
@@ -0,0 +1,263 @@
...
+    edges_y: np.ndarray(dtype=float)
+        edges of histogram in meters
+    speed_thresh: float
+        in meters. Default = 3.0 cm/s
+
+    Returns
+    -------
+    """
+
+    if len(velocity) == 0:
```
bendichter, line 118:
```suggestion
    if velocity is not None:
```

nwbwidgets/analysis/placefields.py
```diff
@@ -0,0 +1,263 @@
...
+        is_running = compute_speed(pos, pos_tt) > speed_thresh
+    else:
+        is_running = velocity > speed_thresh
```
bendichter, line 121: velocity /= speed

nwbwidgets/analysis/placefields.py
```diff
@@ -0,0 +1,263 @@
...
+    filtered_firing_rate = gaussian_filter(firing_rate, gaussian_sd / pixel_width)
+
+    # filter occupancy to create a mask so non-explored regions are nan'ed
+    filtered_occupancy = gaussian_filter(occupancy, gaussian_sd / pixel_width / 8)
+    filtered_firing_rate[filtered_occupancy.astype('bool') < .00001] = np.nan
+
+    return occupancy, filtered_firing_rate, [edges_x, edges_y]
+
+
+def compute_1d_occupancy(pos, pos_tt, spatial_bins, sampling_rate, speed_thresh=0.03):
```
bendichter, line 198: take optional velocity here too

nwbwidgets/placefield.py
```diff
@@ -0,0 +1,321 @@
...
+        return PlaceField_1D_Widget(spatial_series)
+    else:
+        print('Spatial series exceeds dimensionality for visualization')
+        return
+
+
+# Put widget rendering here
+class PlaceFieldWidget(widgets.HBox):
+
+    def __init__(self, spatial_series: pynwb.behavior.SpatialSeries, **kwargs):
```
bendichter, line 66:
```suggestion
    def __init__(self, spatial_series: pynwb.behavior.SpatialSeries, velocity: pynwb.TimeSeries = None, **kwargs):
```

nwbwidgets/placefield.py
```diff
@@ -0,0 +1,321 @@
...
+        super().__init__()
+
+
+        if hasattr(spatial_series.get_ancestor('NWBFile'),'velocity'):
+            velocity = spatial_series.get_ancestor('NWBFile').velocity
+        else:
+            velocity = []
```
bendichter, line 73: We won't be able to find velocity like this. This only really works for units and trials

nwbwidgets/placefield.py
```diff
@@ -0,0 +1,321 @@
...
+
+        self.units = spatial_series.get_ancestor('NWBFile').units
+        self.pos_tt = get_timeseries_tt(spatial_series)
+
```
bendichter, line 77:
```suggestion
        self.velocity = velocity
```

nwbwidgets/placefield.py
```diff
@@ -0,0 +1,321 @@
...
+        ]
+
+    def do_rate_map(self, index=0, speed_thresh=0.03, gaussian_sd=0.0184, use_velocity=False, velocity=[]):
+        tmin = min(self.pos_tt)
+        tmax = max(self.pos_tt)
+
+        spikes = get_spike_times(self.units, index, [tmin, tmax])
+        if use_velocity == False:
+            occupancy, filtered_firing_rate, [edges_x, edges_y] = compute_2d_firing_rate(
+                self.pos, self.pos_tt, spikes, self.pixel_width, speed_thresh=speed_thresh, gaussian_sd=gaussian_sd)
```
bendichter, line 121:
```suggestion
                self.pos, self.pos_tt, spikes, self.pixel_width, speed_thresh=speed_thresh, gaussian_sd=gaussian_sd, velocity=self.velocity)
```

nwbwidgets/placefield.py
```diff
@@ -0,0 +1,321 @@
...
+        bft_speed = BoundedFloatText(value=0.03, min=0, max=99999, description='speed threshold (cm/s)')
+        dd_unit_select = Dropdown(options=np.arange(len(self.units)), description='unit')
+        cb_velocity = Checkbox(value=False, description='use velocity', indent=False)
+
+        self.controls = dict(
+            gaussian_sd=bft_gaussian,
+            speed_thresh=bft_speed,
+            index=dd_unit_select,
+            use_velocity=cb_velocity,
+            velocity=fixed(velocity)
```
bendichter, line 99:
```suggestion
```

bendichter: `fixed` can take a long time for some reason. Try to use class variables or `functools.partial` instead

nwbwidgets/placefield.py
```diff
@@ -0,0 +1,321 @@
...
+            widgets.VBox([
+                bft_gaussian,
+                bft_speed,
+                dd_unit_select,
+                cb_velocity
+            ]),
+            vis2widget(out_fig)
+        ]
+
+    def do_rate_map(self, index=0, speed_thresh=0.03, gaussian_sd=0.0184, use_velocity=False, velocity=[]):
```
bendichter, line 114:
```suggestion
    def do_rate_map(self, index=0, speed_thresh=0.03, gaussian_sd=0.0184, use_velocity=False):
```

nwbwidgets/placefield.py
```diff
@@ -0,0 +1,321 @@
...
+        cb_velocity = Checkbox(value=False, description='use velocity', indent=False)
+
+        self.controls = dict(
+            gaussian_sd=bft_gaussian,
+            speed_thresh=bft_speed,
+            index=dd_unit_select,
+            use_velocity=cb_velocity,
+            velocity=fixed(velocity)
+        )
+
```
bendichter, line 101: (left as an example. Not a suggestion here)
```python
from functools import partial

plot_func = partial(self.do_rate_map, velocity=self.velocity)
```

nwbwidgets/placefield.py
```diff
@@ -0,0 +1,321 @@
...
+        tmax = max(self.pos_tt)
+
+        spikes = get_spike_times(self.units, index, [tmin, tmax])
+        if use_velocity == False:
+            occupancy, filtered_firing_rate, [edges_x, edges_y] = compute_2d_firing_rate(
+                self.pos, self.pos_tt, spikes, self.pixel_width, speed_thresh=speed_thresh, gaussian_sd=gaussian_sd)
+        else:
+            occupancy, filtered_firing_rate, [edges_x, edges_y] = compute_2d_firing_rate(
+                self.pos, self.pos_tt, spikes, self.pixel_width, speed_thresh=speed_thresh, gaussian_sd=gaussian_sd,
+                velocity=velocity)
```
bendichter, line 125:
```suggestion
                velocity=self.velocity)
```

nwbwidgets/placefield.py
```diff
@@ -0,0 +1,321 @@
...
+    def do_1d_rate_map(self, units_window=None, order=None, group_inds=None, labels=None, normalize=False,
+                       collapsed=False, gaussian_sd=0.0557, spatial_bin_len=0.0168, **kwargs):
+        tmin = min(self.pos_tt)
+        tmax = max(self.pos_tt)
+
+        print('Order is {}'.format(order))
+        print('Units_window is {}'.format(units_window))
+        print('Group_inds is {}'.format(group_inds))
+        print('Labels is {}'.format(labels))
+        if order.any() == None:
```
bendichter, line 205:
```suggestion
        if order is None:
```

nwbwidgets/placefield.py
```diff
@@ -0,0 +1,271 @@
...
+                continue
+            spikes = get_spike_times(self.units, ind, [tmin, tmax])
+            xx, _, all_unit_firing_rate[firing_rate_ind] = compute_linear_firing_rate(self.pos, self.pos_tt, spikes,
+                                                                                      gaussian_sd=gaussian_sd,
+                                                                                      spatial_bin_len=spatial_bin_len,
+                                                                                      velocity=self.velocity)
+            firing_rate_ind += 1
+
+
+        fig, ax = plt.subplots()
```
bendichter, line 181:
```suggestion
        fig, ax = plt.subplots(figsize=(7,7))
```

.gitignore
```diff
@@ -129,3 +129,6 @@ dmypy.json
 
 # nwb
 **.nwb
+nwbwidgets/controllers/group_and_sort_controllers_multi_select.py
+nwbwidgets/placefield_group_and_sort.py
+nwbwidgets/towers_task_placefield.py
```
bendichter, line 134: @MRScheid what are these files? Why are you gitignoring them?

PR #127, opened 2020-11-01
add spectrum widget

Saksham20:
added widget to handle ndx-spectrum `Spectrum` datatype
could you try testing : 
```python
from pynwb import NWBHDF5IO
from nwbwidgets import nwb2widget
from IPython.display import display

io = NWBHDF5IO(r"ibl_eid.nwb", mode='r', load_namespaces=True)
nwb = io.read()
a=nwb2widget(nwb)
```
ibl_eid.nwb here https://drive.google.com/file/d/1MF3uitbMxvxpQtGYqE-gr1eNrA2AZc_i/view?usp=sharing
dataset: processing>ephys>_iblqc_ephysSpectralDensity

### DIFF
--- a/nwbwidgets/spectrum.py
+++ b/nwbwidgets/spectrum.py
@@ -54,23 +54,34 @@ def get_spectrum_figure(spectrum,channel_no, freqs):
     else:
         range_ = channel_no[0]-1
     if 'power' in spectrum.fields and 'phase' in spectrum.fields:
+        power_data = np.asarray(spectrum.power)
+        if len(power_data.shape)==1:
+            power_data = power_data[:, np.newaxis]
+        phase_data = np.asarray(spectrum.phase)
+        if len(phase_data.shape)==1:
+            phase_data = phase_data[:, np.newaxis]
         fig, axs = plt.subplots(2, 1, sharex=True)
-        axs[0].semilogy(np.asarray(spectrum.frequencies)[start_id:end_id],
-                        np.asarray(spectrum.power)[start_id:end_id,range_])
+        axs[0].semilogy(all_freqs[start_id:end_id],
+                        power_data[start_id:end_id,range_])
         axs[0].set_ylabel('Power')
-        axs[1].plot(np.asarray(spectrum.frequencies)[start_id:end_id],
-                    np.asarray(spectrum.power)[start_id:end_id,range_])
+        axs[1].plot(all_freqs[start_id:end_id],
+                    phase_data[start_id:end_id,range_])
         axs[1].set_ylabel('phase')
     elif 'power' in spectrum.fields:
+        power_data = np.asarray(spectrum.power)
+        if len(power_data.shape) == 1:
+            power_data = power_data[:, np.newaxis]
         fig, ax = plt.subplots()
         ax.set_xlabel('frequency')
-        
-        ax.semilogy(np.asarray(spectrum.frequencies)[start_id:end_id],
-                    np.asarray(spectrum.power)[start_id:end_id,range_])
+        ax.semilogy(all_freqs[start_id:end_id],
+                    power_data[start_id:end_id,range_])
         ax.set_ylabel('Power')
     elif 'phase' in spectrum.fields:
+        phase_data = np.asarray(spectrum.phase)
+        if len(phase_data.shape) == 1:
+            phase_data = phase_data[:, np.newaxis]
         fig, ax = plt.subplots()
-        ax.plot(np.asarray(spectrum.frequencies)[start_id:end_id],
-                np.asarray(spectrum.power)[start_id:end_id,range_])
+        ax.plot(all_freqs[start_id:end_id],
+                phase_data[start_id:end_id,range_])
         ax.set_ylabel('phase')
         ax.set_xlabel('frequency')
--- a/nwbwidgets/spectrum.py
+++ b/nwbwidgets/spectrum.py
@@ -37,31 +37,31 @@ def show_spectrum(node: Spectrum, **kwargs) -> widgets.Widget:
     out = widgets.Output()
     with out:
         widgets.interact(
-            lambda channel_no, frequency_range: plot_spectrum_figure(node, channel_no, frequency_range),
-            channel_no=channel_slider,
+            lambda channel_range, frequency_range: plot_spectrum_figure(node, channel_range, frequency_range),
+            channel_range=channel_slider,
             frequency_range=freq_slider)
     return out
 
 
-def plot_spectrum_figure(spectrum, channel_no, freqs):
+def plot_spectrum_figure(spectrum, channel_nos, frequency_nos):
     """
     Plot power vs frequencies and/or phase vs frequencies.
     Parameters
     ----------
     spectrum: Spectrum
-    channel_no: tuple
+    channel_nos: tuple
         Input from the channel range slider widget: (channel_no start, channel_no end)
-    freqs: tuple
+    frequency_nos: tuple
         Input from frequency range slider widget: (freq start, freq end)
     """
     check_spectrum(spectrum)
     all_freqs = np.asarray(spectrum.frequencies)
-    start_id = (np.abs(freqs[0] - all_freqs)).argmin()
-    end_id = (np.abs(freqs[1] - all_freqs)).argmin()
-    if channel_no[0] != channel_no[1]:
-        range_ = range(channel_no[0] - 1, channel_no[1] - 1)
+    start_id = (np.abs(frequency_nos[0] - all_freqs)).argmin()
+    end_id = (np.abs(frequency_nos[1] - all_freqs)).argmin()
+    if channel_nos[0] != channel_nos[1]:
+        range_ = range(channel_nos[0] - 1, channel_nos[1] - 1)
     else:
-        range_ = channel_no[0] - 1
+        range_ = channel_nos[0] - 1
     if 'power' in spectrum.fields:
         power_data = np.asarray(spectrum.power)
         if len(power_data.shape) == 1:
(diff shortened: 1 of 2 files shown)


nwbwidgets/spectrum.py
```diff
@@ -0,0 +1,59 @@
...
+        return out
+    else:
+        warnings.warn('neurodatatype not of type: Spectrum')
+
+
+def _data_to_array(data):
+    if isinstance(data, Dataset):
+        return data[()]
+    else:
+        return np.array(data)
```
bendichter, line 31: `np.asarray` may work instead of this function

nwbwidgets/spectrum.py
```diff
@@ -0,0 +1,59 @@
...
+from collections.abc import Iterable
+from h5py import Dataset
+from ipywidgets import widgets
+import numpy as np
+from ipywidgets.widgets.interaction import show_inline_matplotlib_plots
+
+
+def show_spectrum(node, **kwargs):
+    if isinstance(node, Spectrum):
+        power = _data_to_array(node.power)
```
bendichter, line 13: `np.asarray` may work instead of this function

nwbwidgets/spectrum.py
```diff
@@ -0,0 +1,59 @@
...
+import warnings
+import matplotlib.pyplot as plt
+from collections.abc import Iterable
+from h5py import Dataset
+from ipywidgets import widgets
+import numpy as np
+from ipywidgets.widgets.interaction import show_inline_matplotlib_plots
+
+
+def show_spectrum(node, **kwargs):
```
bendichter, line 11: use typehints here

nwbwidgets/spectrum.py
```diff
@@ -0,0 +1,59 @@
...
+    if isinstance(node, Spectrum):
```
bendichter, line 12: do type checking at the beginning of the function. Should be `assert isinstance(node, Spectrum), 'node must be of type Spectrum'`

nwbwidgets/spectrum.py
```diff
@@ -0,0 +1,59 @@
...
+        power = _data_to_array(node.power)
+        if len(power.shape)==2:
+            no_channels = power.shape[1]
+        else:
+            no_channels = 1
+        out = widgets.Output()
+        with out:
+            widgets.interact(lambda channel_no: sp(node, channel_no), channel_no=(0, no_channels, 1))
+            show_inline_matplotlib_plots()
+        return out
```
bendichter, line 22: use vis2widget

nwbwidgets/spectrum.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def _data_to_array(data):
+    if isinstance(data, Dataset):
+        return data[()]
+    else:
+        return np.array(data)
+
+
+def sp(spectrum,channel_no):
```
bendichter, line 34: please use more descriptive function names

bendichter (changes requested):
good start @Saksham20, could you address these comments?

nwbwidgets/spectrum.py
Saksham20: fixed https://github.com/NeurodataWithoutBorders/nwb-jupyter-widgets/pull/127/commits/b2bea61fd860aa2c7cb1366034e2d87dc1f708b9

nwbwidgets/spectrum.py
Saksham20: I'm also adding a slider with the plot to show only one channel. `vis2widget `is called by the parent function anyway (if the output is anything other than a plot).
One option that I considered was to only output a plot of multi-channel traces of power in one figure, but the plot wasn't very good to look at,so I added a **slider** (using the widgets.interact) to select only one channel to plot the trace for.

To link a `widget.value` to an input arg to a figure I tried two ways: to slip using the widgets.Output() and only returning a widgets.interact(), but in this case the `show_inline_matplotlib_plots()` was not working during rendering. With out that it correctly shows inline but there are two figures instead of one. Can you try it with the nwb file in the main draft comment?

nwbwidgets/spectrum.py
Saksham20: I wanted to raise only a warning if it fails

nwbwidgets/spectrum.py
Saksham20: fixed https://github.com/NeurodataWithoutBorders/nwb-jupyter-widgets/pull/127/commits/b2bea61fd860aa2c7cb1366034e2d87dc1f708b9

nwbwidgets/spectrum.py
Saksham20: done

nwbwidgets/spectrum.py
Saksham20: as an update: there are two sliders now: one for channel range and other for frequency range. The double plot showing up problem is solved but I still have kept this as a `widgets.Output()`. vis2widget would not work since it checks for the function returning a plt.plot() object

nwbwidgets/spectrum.py
bendichter: ok then do that first

bendichter: This is a function that requires a certain data type. If it is not given that data type, it should throw an appropriate error, not return None

nwbwidgets/view.py
```diff
@@ -1,4 +1,4 @@
-from collections import OrderedDict
+from collections import OrderedDict, abc
```
bendichter, line 1: why are you importing this?

nwbwidgets/spectrum.py
```diff
@@ -0,0 +1,88 @@
...
+        ax.set_ylabel('Power')
+    elif 'phase' in spectrum.fields:
+        phase_data = np.asarray(spectrum.phase)
+        if len(phase_data.shape) == 1:
+            phase_data = phase_data[:, np.newaxis]
+        fig, ax = plt.subplots()
+        ax.plot(all_freqs[start_id:end_id],
+                phase_data[start_id:end_id,range_])
+        ax.set_ylabel('phase')
+        ax.set_xlabel('frequency')
```
bendichter, line 88: Lots of repeated code here. Can you clean this up?

nwbwidgets/spectrum.py
```diff
@@ -0,0 +1,88 @@
...
+            widgets.interact(
+                lambda channel_no, frequency_range: get_spectrum_figure(node, channel_no, frequency_range),
+                channel_no=channel_slider,
+                frequency_range=freq_slider)
+        return out
+    else:
+        warnings.warn('neurodatatype not of type: Spectrum')
+
+
+def get_spectrum_figure(spectrum,channel_no, freqs):
```
bendichter, line 49: Use docstring to explain what these args should be

nwbwidgets/spectrum.py
```diff
@@ -0,0 +1,88 @@
...
+            readout=True,
+            readout_format='d',
+        )
+        out = widgets.Output()
+        with out:
+            widgets.interact(
+                lambda channel_no, frequency_range: get_spectrum_figure(node, channel_no, frequency_range),
+                channel_no=channel_slider,
+                frequency_range=freq_slider)
+        return out
```
bendichter, line 44: please use vis2widget

nwbwidgets/spectrum.py
bendichter: please use utils.widget.interactive_output like it is done here: https://github.com/NeurodataWithoutBorders/nwb-jupyter-widgets/blob/81b8b4f977dc7d40dc4b0f188d783b94f224a4fc/nwbwidgets/misc.py#L110-L117

nwbwidgets/spectrum.py
Saksham20: it takes input from the slider, which can only be a continuous range

bendichter: then wrap it in another function that takes the slider input and changes it to a list of ints

nwbwidgets/spectrum.py
Saksham20: but will it not be called anyway? https://github.com/NeurodataWithoutBorders/nwb-jupyter-widgets/blob/master/nwbwidgets/base.py#L204
why use it twice?

nwbwidgets/spectrum.py
Saksham20: line 54 is doing that

bendichter: ok, can you put a docstring in though?

nwbwidgets/spectrum.py
bendichter: I'd still like you to do this. Cutting out early allows you to un-indent the entire rest of the function.

Saksham20: okay makes sense, fixed https://github.com/NeurodataWithoutBorders/nwb-jupyter-widgets/pull/127/commits/3b9f7357ecccfbba24fa8fc10d42c2f81c205a6c

bendichter:
one more thing: could you in run `PyCharm>Code>Reformat Code`? This will take care of little formatting things like import order and white space. Could you make a habit of doing this?

nwbwidgets/spectrum.py
```diff
@@ -0,0 +1,82 @@
...
+    out = widgets.Output()
+    with out:
+        widgets.interact(
+            lambda channel_no, frequency_range: get_spectrum_figure(node, channel_no, frequency_range),
+            channel_no=channel_slider,
+            frequency_range=freq_slider)
+    return out
+
+
+def get_spectrum_figure(spectrum,channel_no, freqs):
```
bendichter, line 46: Could you add a docstring here?

Saksham20: done

Saksham20:
> one more thing: could you in run `PyCharm>Code>Reformat Code`? This will take care of little formatting things like import order and white space. Could you make a habit of doing this?

okay done

PR #128, opened 2020-12-02
Add custom timeseries widget 

weiglszonja:
Use a different timeseries widget when plotting "Velocity"


bendichter:
Can you implement these changes in the tank-lab-to-nwb repo?

PR #133, opened 2021-02-16
prevent image_mask from loading into memory

### LINKED ISSUE #135: visualizing ROIs from PlaneSegmenation is time and memory intensive for large datasets
I am working with a large dataset: plane_segmentation['image_mask'].data.shape = (1729, 512, 796).

Viewing the ROis from widgets takes over 120 seconds on my machine and is memory intensive (~6GB). 

Is there a possibility to add the functionality to show a small number of accepted/rejected rois in the default plot (maybe using a slider)?

Saksham20:
to label all the ROIs, all the data of `planesegmentation `is loaded into memory as dataframe. For large datasets, its very memory intensive. I think its unnecessary to load 'image_masks' and 'pixel_masks' as they are not needed as labels of each ROI in the plot. 
fixes #135

### DIFF
--- a/nwbwidgets/ophys.py
+++ b/nwbwidgets/ophys.py
@@ -169,18 +169,20 @@ def show_plane_segmentation_2d(
         )
 
     aux_leg = []
-    all_hover = df_to_hover_text(plane_seg.to_dataframe())
-
+    import pandas as pd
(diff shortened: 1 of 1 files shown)


bendichter:
Thanks, this is really helpful!

bendichter (approved):

PR #134, opened 2021-02-16
load roi_table_region data into mem

Saksham20:
looping over all the ROIs and checking if each roi's id exists in a `roi_response_series.rois ` (**DynamicTableRegion** linked to the rois table) is slow since the roi_response_series.rois.data is on disc. 
Loading into memory as array makes the search fast. I dont think we would run into any memory issues though since the roi table region only has one column with a subset number of rois. 

I faced this issue when I tried to visualize the roi_response_series for two datasets 1) shape=(100, Time)  2) shape=(1900,Time). 
I was taking a long time (190+ sec) to render the default plot of widget for dataset no.2, while dataset no.1 was 2-3 secs. This change makes the dataset 1 plot also render quickly (2-3 sec)


bendichter:
nice work tracking this down!

PR #141, opened 2021-02-20
Brunton brains

MRScheid:
Here is a screenshot of the widget currently: 

image


MRScheid:
For some reason whenever I toggle the new "Color by:" radio button, the handler is called twice.  I can't figure out why this is.

bendichter:
Could you make coloring by electrode group a 3rd option? Also ideally the additional options would be automatically determined by the columns of the electrodes table. That would make this generalize to other groups

MRScheid:
> 
> 
> Could you make coloring by electrode group a 3rd option? Also ideally the additional options would be automatically determined by the columns of the electrodes table. That would make this generalize to other groups


@bendichter took care of this in the last commit.  Here's a screenshot of the current layout:
image

bendichter:
Awesome! Could you have the program automatically determine whether the column is continuous or categorical and only show the colorbar if it is continuous? Also, could you use a colormap other than jet?

MRScheid:
image

Done and done.  Using viridis now.

bendichter:
Looks good! Can you add a test?

bendichter (approved):

PR #142, opened 2021-02-22
progress bar add

Saksham20:
this PR adds a progress bar while loading image_masks of planesegmentation table.

### DIFF
--- a/nwbwidgets/ophys.py
+++ b/nwbwidgets/ophys.py
@@ -21,6 +21,7 @@ from .timeseries import BaseGroupedTraceWidget
 from .utils.cmaps import linear_transfer_function
 from .utils.dynamictable import infer_categorical_columns
 from .utils.functional import MemoizeMutable
+from .controllers import ProgressBar
 
 color_wheel = ["red", "blue", "green", "black", "magenta", "yellow"]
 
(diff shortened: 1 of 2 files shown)


nwbwidgets/ophys.py
```diff
@@ -160,20 +161,27 @@ def __init__(self, plane_seg: PlaneSegmentation, color_wheel=color_wheel, **kwar
...
         self.plane_seg = plane_seg
         self.color_wheel = color_wheel
-
+        self.progress_bar = ProgressBar()
+        self.button = widgets.Button(description='display Image Masks')
+        self.children = [self.progress_bar.container, self.button]
+        self.button.on_click(self.on_button_click)
+        self.kwargs = kwargs
+
+    def on_button_click(self, b):
```
Saksham20, line 170: created a button serves two purposes:
1) without this, the progress bar does not show up since jupyter notebook does not show the bar until the figure is populated with all the image_masks (which is what the progress bar tracks).
2) to prevent the notebook from freezing if the user by mistake clicks on the planesegmentation dropdown.

bendichter:
Good start! However when I test this in my computer the button and progress bar are only working on the first run, and not on subsequent runs when other items from the dropdown menu are selected. I think it would be the other way around- it should run automatically on the first run and then should require a button press for subsequent runs.

Saksham20:
> Good start! However when I test this in my computer the button and progress bar are only working on the first run, and not on subsequent runs when other items from the dropdown menu are selected. I think it would be the other way around- it should run automatically on the first run and then should require a button press for subsequent runs.

I was actually wanting it to show the progress bar and the button only in the first run when it loads all the image_masks (which takes time). Once that is set, the changes that take place when choosing from the drop down menu dont have a progress bar+button.

Saksham20:
I dont see a way to have a progress bar when selecting the options from the drop down menu because: this line [here](https://github.com/NeurodataWithoutBorders/nwb-jupyter-widgets/pull/142/files/97d42a3574195b28505899f3137f4be8ea48ef15#diff-75f6ccd6223435213e08bf8d1446ba112a7046b341faf360fc3af1000712a229R202)(`fig.batch_update`)
is rendering only when all iterations are complete so when I put a progress bar to update within this loop, it does not update in real time but only after all the iterations are complete.

bendichter:
OK, I understand. Then can we remove the progress bar once it is finished?

Saksham20:
> OK, I understand. Then can we remove the progress bar once it is finished?

now the button and the bar should go away

PR #143, opened 2021-02-23
Progressbar ophys

Saksham20:
the old merged PR #142 , after last commit https://github.com/NeurodataWithoutBorders/nwb-jupyter-widgets/pull/142/commits/71340222d283fd32e72d0e499a16ed96b8cdb9d7: in which the button+bar were grouped in a HBox widget, ends up removing the drop down selection widget also. This PR fixes that.

### DIFF
--- a/nwbwidgets/ophys.py
+++ b/nwbwidgets/ophys.py
@@ -196,7 +196,7 @@ class PlaneSegmentation2DWidget(widgets.VBox):
             self.children += (
                 self.show_plane_segmentation_2d(color_by=None, **self.kwargs),
             )
-        self.children = self.children[2:]
+        self.children = self.children[1:]
 
     def update_fig(self, color_by):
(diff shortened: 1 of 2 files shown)


nwbwidgets/ophys.py
```diff
@@ -162,8 +162,8 @@ def __init__(self, plane_seg: PlaneSegmentation, color_wheel=color_wheel, **kwar
         self.plane_seg = plane_seg
         self.color_wheel = color_wheel
         self.progress_bar = ProgressBar()
-        self.button = widgets.Button(description='display Image Masks')
-        self.children = [self.progress_bar.container, self.button]
+        self.button = widgets.Button(description="Display ROIs")
+        self.children = [widgets.HBox([self.button, self.progress_bar.container])]
```
Saksham20, line 166: Grouping then in a HBox clips the progress bar. Does that happen with you too? 
Looks like : 
image

PR #148, opened 2021-04-06
tuning curve widget

### LINKED ISSUE #166: MatplotlibDeprecationWarning:  is_first_col and is_last_row
image

### LINKED ISSUE #170: Very slow performance with ipywidgets.fixed()
The function `ipywidgets.fixed()` slows significantly the widgets for some pynwb objects, depending on the size of the object it won’t even finish rendering it. I found this in some big files like `sub-699733573_ses-715093703.nwb` from Allen dataset and with time intervals with many intervals.

What I did to solve it was to allow a fixed dictionary to be passed as an optional argument to `interactive_output()`. It seems to have solved the issue. Check 56ab45b

luiztauffer:
- [x] Create UnitsAndTrialsControllerWidget
- [x] Create TuningCurveWidget for nwb files with Units and Trials tables
- [x] Have a combined widget with RasterGridWidget and TuningCurveWidget 
- [x] Fix #166 
- [x] Fix #170 
- [x] Start and End controllers as FloatText widgets (from negative to positive values), allowing for more flexible control of calculation regions around the reference trial marks


bendichter:
thanks @luiztauffer can you include a screenshot please?

nwbwidgets/misc.py
```diff
@@ -1078,3 +1081,126 @@ def show_session_raster_plotly(
...
+        avg_rates = []
+        for v1 in var1_classes:
+            indexes = np.where(self.trials[rows_label][:]==v1)[0].tolist()
+            data = align_by_time_intervals(
+                units=self.units,
+                index=index,
+                intervals=self.trials,
+                start_label=start_label,
+                stop_label=start_label,
+                before=0.0,
```
bendichter, line 1182: I'd rather not hard-code `before`

nwbwidgets/misc.py
```diff
@@ -1078,3 +1081,126 @@ def show_session_raster_plotly(
...
+    ):
+
+        if rows_label is None:
+            return widgets.HTML("Select at least one variable")
+
+        var1_classes = np.unique(self.trials[rows_label][:]).tolist()
+
+        avg_rates = []
+        for v1 in var1_classes:
+            indexes = np.where(self.trials[rows_label][:]==v1)[0].tolist()
```
bendichter, line 1175: it looks like `cols_labels `is unused

bendichter (changes requested):

luiztauffer:
@bendichter thanks for the comments, this is how it looks like so far

image

nwbwidgets/misc.py
luiztauffer: for now... next step is to make it 2d

nwbwidgets/misc.py
luiztauffer: would anyone use the rate before stimuli to generate tuning curves?

bendichter: Yes, if it's a motor task you'll expect activity before the action. Also, for sensory stimuli you might want the summation window to start after the stimulus onset

nwbwidgets/misc.py
```diff
@@ -1078,3 +1081,126 @@ def show_session_raster_plotly(
...
+
+        unit_controller = widgets.Dropdown(
+            options=range(len(units["spike_times"].data)),
+            value=unit_index,
+            description="unit",
+        )
+
+        after_slider = widgets.FloatSlider(
+            1.0, min=0, max=5.0, description="window (s)", continuous_update=False
+        )
```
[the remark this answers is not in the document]
bendichter: or just two float selectors which control the start and stop of the summation window

luiztauffer:
@bendichter 

Peek 2021-04-29 17-23

bendichter:
@luiztauffer this looks really great! What happens if you select something for cols, but not rows? Maybe it makes sense to have cols be inactive when rows is None, and activate it when a row value is selected.

bendichter:
@luiztauffer 

You aren't quite handling intervals correctly. You allow input of a TimeIntervals object (this is how I would prefer) and if this is not supplied, you auto-populate it with nwb.intervals, which is not of type TimeIntervals. Instead, can you make the interval selection part its own widget? See how this was done in 

https://github.com/NeurodataWithoutBorders/nwb-jupyter-widgets/blob/c72454e6c30dda33c7313f86cf9ea66d9de28f3e/nwbwidgets/allen.py#L22-L50 

Happy to jump on a call and discuss if this is confusing.

bendichter:
Also, can you please create a test?

luiztauffer:
> @luiztauffer
> 
> You aren't quite handling intervals correctly. You allow input of a TimeIntervals object (this is how I would prefer) and if this is not supplied, you auto-populate it with nwb.intervals, which is not of type TimeIntervals. Instead, can you use `TimeIntervalsSelector`? When you do this, can you move `TimeIntervalsSelector` from `allen.py` into `base.py` and adjust the variable names within the class to be more generic?
> 
> https://github.com/NeurodataWithoutBorders/nwb-jupyter-widgets/blob/c72454e6c30dda33c7313f86cf9ea66d9de28f3e/nwbwidgets/allen.py#L22-L50
> 
> Happy to jump on a call and discuss if this is confusing.

@bendichter 
sure, I'm on it!

1. Should we change the default behavior of `default_neurodata_vis_spec` when using classes such as `misc.PSTHWidget` and `misc.RasterGridWidget`, to instead use classes such as these ones you did for Allen, eg `AllenRasterGridWidget`? That is, to make them already come with a TimeIntervalSelector controller by default?

2. when we call `misc.PSTHWidget` and `misc.RasterGridWidget`, if no `TimeIntervals` is passed as input they search for `self.units.get_ancestor("NWBFile").trials`, and that will not get intervals not defined as trials. Should we change this to fetch the first item in `self.units.get_ancestor("NWBFile").intervals` instead?

bendichter:
> 1. Should we change the default behavior of default_neurodata_vis_spec when using classes such as misc.PSTHWidget and misc.RasterGridWidget, to instead use classes such as these ones you did for Allen, eg AllenRasterGridWidget? That is, to make them already come with a TimeIntervalSelector controller by default?

No, that was a special case. It wouldn't make sense for most labs because they just use the trials table.

> 2. when we call misc.PSTHWidget and misc.RasterGridWidget, if no TimeIntervals is passed as input they search for self.units.get_ancestor("NWBFile").trials, and that will not get intervals not defined as trials. Should we change this to fetch the first item in self.units.get_ancestor("NWBFile").intervals instead?

No, let's leave it as defaulting to `trials`. Again, most labs just use that. I just want to make sure we can customize it appropriately in a few lines of code.

bendichter:
@luiztauffer this looks great. Let me know once you test everything and I'll merge

nwbwidgets/base.py
```diff
@@ -371,6 +371,39 @@ def row_to_hover_text(row):
     return "<br>".join(text_rows)
 
 
+class TimeIntervalsSelector(widgets.VBox):
+    InnerWidget = None
+
+    def __init__(self, units, **kwargs):
+        super().__init__()
+        self.units = units
```
bendichter, line 379: Can you generalize this to work for any input data, not just units?

nwbwidgets/misc.py
```diff
@@ -1087,3 +1092,196 @@ def show_session_raster_plotly(
...
+        Gets triggered when self.rows_controller changes. Updates other dropdown options.
+        """
+        if change['new'] is None:
+            self.cols_controller.disabled = True
+            self.cols_controller.value = None
+        else:
+            self.cols_controller.disabled = False
+
+
+def draw_tuning_curve(
```
bendichter, line 1196: Could you separate this out into 2 functions, routed by whether col_label is None?

nwbwidgets/base.py
bendichter: maybe instead of "units" it could take "input_data" which is a list of objects that go into the inner widget

nwbwidgets/misc.py
luiztauffer: done

luiztauffer:
@bendichter 
`TimeIntervalsSelector` is no longer constrained to receive Units as input_data. It still requires that the method input_data.get_ancestor(‘NWBFile’) will work for the passed input_data though.
To make it accept a list of objects would require changes in `base.nwb2widget()`, where the specs are instantiated. This might affect other widgets, so maybe we bring it on another PR?

The function `ipywidgets.fixed()` slows significantly the widgets for some pynwb objects, depending on the size of the object it won’t even finish rendering it. I found this in some big files like `sub-699733573_ses-715093703.nwb` from Allen dataset and time intervals with many intervals. It was not the first time I met this problem.
What I did to solve it was to allow a `fixed` dictionary to be passed as an optional argument to `interactive_output`. It seems to have solved the issue.

luiztauffer:
For now, that's how it looks like:

Peek 2021-06-15 16-36

luiztauffer (commented):
comments addressed

nwbwidgets/base.py
```diff
@@ -371,6 +371,48 @@ def row_to_hover_text(row):
...
+        )
+        self.children = [self.stimulus_type_dd, inner_widget]
+
+    def stimulus_type_dd_callback(self, change):
+        self.children = [self.stimulus_type_dd, widgets.HTML("Rendering...")]
+        trials = self.intervals_tables[self.stimulus_type_dd.value]
+        inner_widget = self.InnerWidget(
+            input_data=self.input_data, 
+            trials=trials, 
+            **self.kwargs
```
bendichter, line 411: ah, maybe everything can be accomplished with self.kwargs, and we don't need self.input_data. What do you think?

nwbwidgets/misc.py
```diff
@@ -269,14 +271,14 @@ def control_plot(x0, x1, ch0, ch1):
 class PSTHWidget(widgets.VBox):
     def __init__(
         self,
-        units: Units,
```
bendichter, line 272: Oh no, this isn't what I meant, but I can see why you thought that

luiztauffer:
@bendichter `misc.RasterGridWidget` and `misc.TuningCurveWidget` are independent components now and can be put together in a `misc.TuningCurveExtendedWidget`. They all require a `UnitsAndTrialsControllerWidget`

luiztauffer:
@bendichter the FloatSliders controllers were substituted for FloatText boxes and tests are all passing, let's merge it?

bendichter (approved):

bendichter:
@luiztauffer lgtm

PR #149, opened 2021-04-07
include custom widget

luiztauffer:
This allows user-created widgets to be included in the `neurodata_vis_spec`, can be used like this:

```python
import pynwb
from pynwb import NWBFile, NWBHDF5IO
from nwbwidgets import nwb2widget
from mywidgets import TunningCurvesWidget,

io = NWBHDF5IO('file.nwb', 'r')
nwbfile = io.read()

# include custom widgets
include_widgets = [{
    'widget': TunningCurvesWidget,
    'label': 'Tuning Curve',
    'pynwb_class': pynwb.misc.Units
}]

nwb2widget(node=nwbfile, include_widgets=include_widgets)
```

For now it only works for items which are `OrderedDict`. Maybe we would like to extend most items to be `OrderedDict`? This would in principle just nest each item in a single Tab, but allows for the possibility of easily adding customized widgets


nwbwidgets/view.py
```diff
@@ -82,5 +83,31 @@ def show_dynamic_table(node, **kwargs) -> widgets.Widget:
...
+
     return base.nwb2widget(node, neurodata_vis_spec)
+
+
+def include_widget(neurodata_vis_spec, widget, label, pynwb_class):
+    if pynwb_class in neurodata_vis_spec:
+        neurodata_vis_spec[pynwb_class].update({
+            label: widget
+        })
+    return neurodata_vis_spec
```
bendichter, line 113: @luiztauffer cool idea. Could you also handle if the neurodata_type is not present? Something like:

```suggestion
def include_widget(neurodata_vis_spec, widget, label, pynwb_class):
    if pynwb_class in neurodata_vis_spec:
        neurodata_vis_spec[pynwb_class].update({label: widget})
    else:
        neurodata_vis_spec[pynwb_class] = {label: widget}
    return neurodata_vis_spec
```

bendichter:
@luiztauffer see edits to my comment. You may have to test it a bit. I'm not 100% sure it will work as I wrote it

bendichter:
Now that I am thinking about it, maybe the values of `neurodata_vis_spec` should *always* be `OrderedDict`s, and we would only present it as `lazy_tabs` when the length is > 1. What do you think?

luiztauffer:
@bendichter I agree, at least with most of the items. It makes adding custom widgets easy

PR #152, opened 2021-04-24
create widget for trial-aligned timeseries

bendichter:
@Saksham20 and @MRScheid both of you are working on projects that need a way to trial-align timeseries data. What do you think of this?

* fix gas for when control_limit is false
* align_by_times now takes starts and durations
* add AlignMultiTraceTimeSeriesByTrials
* add TrialAlignedSpacialSeries

image


MRScheid:
@bendichter it looks good.  The only thing I can think of is having two more dropdowns--one to select any arbitrary joint position in the nwb and another to select any behavioral event (i.e. the coarse events) to align to.  Though I'm thinking the coarse event timing might not be aligned to any specific behavior, e.g. the eating label is probably not aligned to when the subject had their hand at their mouth, for instance, so maybe that's not worth implementing just yet.

Saksham20:
@bendichter the timeseries in my dataset is uneven with a constant rate though. Every trial as some varying amount of data points, so the timeseries has discontinuity. 
So the `utils/timeseries.align_by_times()` was creating a np.array out of a ragged array as the `idx_stop-idx_start` is not same for every trial. To deal with this, I've proposed some changes to the utils/timeseries.py file where `align_by_times()` makes the trialized data into same length chunks with length of the smallest trial length. I think this should only have a differed output for such rare cases of discontinuous timestamps.  
Also, the changes in the nwbwidgets/timeseries.py reflect this too; I've create a temporary timeseries object with data as the timestamps to get the trialized timestamps which then I mean over the trial dimension to get the mean `tt` for the x axis range.

Saksham20:
@bendichter I've pushed some changes @https://github.com/NeurodataWithoutBorders/nwb-jupyter-widgets/pull/152/commits/f265f987b95fd13b9a3f48a5b2ea37168472ed31 based on our discussion. Briefly:

- `Timeseries.timestamps` are not assumed to be of constant rate. There are separate methods for trial alignment of timeseries when rate or timestamps is present. (_utils/timeseries.py_)
- The trial alignment of timeseries widget has an inheritance structure with child classes for each of the two cases: timeseries with timestamps or rate+starting_time. (_nwbwidgets/timeseries.py_).

nwbwidgets/timeseries.py
```diff
@@ -654,3 +660,267 @@ def __init__(self, time_series_list, widget_class_list, constrain_time_range=Fal
...
+            self.time_series,
+            self.trials,
+            start_label,
+            before,
+            after,
+            traces=index,
+        )
+        starts = np.array(self.trials[start_label][:]) - before
+        time_ts_aligned = align_timestamps_by_trials(self.time_series,starts,duration=before+after)
+        time_ts_aligned = [i-i[0]-before for i in time_ts_aligned]
```
bendichter, line 898: I think this line should go within `align_timestamps_by_trials`

nwbwidgets/utils/timeseries.py
```diff
@@ -137,29 +137,97 @@ def timeseries_time_to_ind(node: TimeSeries, time, ind_min=None, ind_max=None) -
...
+    duration: float
+        duration in seconds
+    Returns
+    -------
+    out: list
+        list: length=(n_trials); list[0]: array, shape=(n_time, ...)
+    """
+    assert timeseries.rate is not None, 'supply timeseries with start_time and rate'
+    ts = np.arange(timeseries.data.shape[0])*1/timeseries.rate + timeseries.starting_time
+    return bisect_array_by_times(timeseries.data, ts, starts, duration, traces)
```
bendichter, line 206: If you are given a starting time and rate, you should calculate the indices of `data` directly, not create a timeseries and bisect it.

bendichter:
looking good! please address comments and fix tests

bendichter:
@Saksham20 thanks! Now if you could just fix the tests we'll be ready to merge :-)

nwbwidgets/utils/timeseries.py
```diff
@@ -56,9 +56,9 @@ def get_timeseries_maxt(node: TimeSeries) -> float:
     if node.timestamps is not None:
         return node.timestamps[-1]
     elif np.isnan(node.starting_time):
-        return len(node.data) / node.rate
+        return (len(node.data)-1) / node.rate
```
Saksham20, line 59: for a timesries: with rate=1 and starting_time=0, len(data) = 10, 
the maxt should be 9 no?

bendichter: yeah, good catch

nwbwidgets/utils/timeseries.py
```diff
@@ -128,7 +128,7 @@ def timeseries_time_to_ind(node: TimeSeries, time, ind_min=None, ind_max=None) -
             kwargs.update(lo=ind_min)
         if ind_max is not None:
             kwargs.update(hi=ind_max)
-        return bisect(node.timestamps, time, **kwargs)
+        return bisect(node.timestamps, time, **kwargs)-1
```
Saksham20, line 131: again,
for timestamps=[0,1,2,3,4,5]
the index at time=5 should be 5 but bisect will return 6 since it will find the location where if 5 is inserted, will keep the list in sorted order.

bendichter: look at method `bisect_left`

Saksham20: okay, i've made a change here: https://github.com/NeurodataWithoutBorders/nwb-jupyter-widgets/pull/152/commits/e2a0abb5be07d46fba48a8fc27027954c703dded

Saksham20 (approved):

Saksham20 (approved):

Saksham20:
@bendichter  PR ready to be merged

bendichter:
@Saksham20 can you please add tests for the new classes here?

Saksham20:
> @Saksham20 can you please add tests for the new classes here?

added tests, which increased the codecov beyond the threshold

bendichter:
@Saksham20 thanks, looks good! I just renamed "behavior_trial_align" to "trial_align_spatial_series" and applied Black formatting.

PR #155, opened 2021-05-04
add range controller to select traces for any timeseries

Saksham20:
add a feature to select the traces to plot for a multi trace `TimeSeries `which does not have any dynamic table linked to it (unlike `ElectricalSeries`)


nwbwidgets/timeseries.py
```diff
@@ -483,6 +487,7 @@ def plot_grouped_traces(
             offsets[-1] + (offsets[-1] - offsets[-2]) / 2,
         )
     if len(order) <= 30:
+        print(offsets, order)
```
bendichter, line 490:
```suggestion
```

PR #156, opened 2021-05-05
generalize multi container interface

bendichter:
@Saksham20, does this change work for you?


nwbwidgets/base.py
```diff
@@ -370,3 +370,17 @@ def row_to_hover_text(row):
         elif isinstance(val, float):
             text_rows.append("{}: {:.3f}".format(key, val))
     return "<br>".join(text_rows)
+
+
+def show_multi_container_interface(
```
Saksham20, line 375: okay, it does show up all the datainterfaces within the multicontainer, but does not show any labels.
So currently, nwbfile.processing['ecephys']['LFP']['A001'/'B001'..]. The accordian for 'processing' directly shows the 'A001','B001'. I think there should be another accordian for LFP
image

nwbwidgets/base.py
```diff
@@ -96,9 +96,9 @@ def show_neurodata_base(
...
 
-def dict2accordion(
-    d: dict, neurodata_vis_spec: dict, **pass_kwargs
-) -> widgets.Accordion:
+def dict2accordion(d: dict, neurodata_vis_spec: dict, **pass_kwargs) -> widgets.Widget:
+    if len(d) == 1:
+        return nwb2widget(list(d.values())[0], neurodata_vis_spec=neurodata_vis_spec)
     children = [widgets.HTML("Rendering...") for _ in d]
     accordion = widgets.Accordion(children=children, selected_index=None)
     for i, label in enumerate(d):
```
Saksham20, line 104:
```suggestion
    for i, label in enumerate(d):
      if hasattr(d[label], "description") and d[label].description:
              accordion.set_title(i, label + ": " + d[label].description)
          else:
              accordion.set_title(i, label)
```
essentially line 109 should be removed. Github does not allow me to write suggestion on that line

Saksham20 (changes requested):

nwbwidgets/base.py
Saksham20: or the other possibility is to leave this as it is, since by making the 109 line deletion in the below [suggestion](https://github.com/NeurodataWithoutBorders/nwb-jupyter-widgets/pull/156#discussion_r627029939) populates the accordian title with the description, which is self-explanatory

bendichter: Oh, that's intentional. If an accordion as only a single option, I just go straight through to that option. So since LFP is the only object within ecephys, I just assume that's the one you'd want to select and forward you on to the contents of LFP

PR #157, opened 2021-05-07
Range controller tseries

Saksham20:
when there are no categorical columns found for `DynamicTable`, then switch to a `time_series controller` to select traces. 
`infer_categorical_columns()` takes array of rows to index into as argument to calculate categorical columns.


nwbwidgets/utils/dynamictable.py
```diff
@@ -1,21 +1,41 @@
...
+                    column_data = []
+                unique_vals = np.unique(column_data)
+                if 1 < len(unique_vals) <= (len(column_data) / 2):
                     unique_vals = [
                         x.decode() if isinstance(x, bytes) else x for x in unique_vals
                     ]  # handle h5py 3.0
-                    categorical_cols[name] = unique_vals
-            except:
-                pass
+                    categorical_cols[name] = np.array(column_data)
```
Saksham20, line 36: I think its equally feasible to return a complete list of col data instead of unique values. These can then be reused by the `GroupAndSortController.get_group_vals(). `

nwbwidgets/utils/dynamictable.py
```diff
@@ -1,21 +1,41 @@
...
+    region = region if region is not None else list(range(len(dynamic_table)))
     for name in dynamic_table.colnames:
         if len(dynamic_table[name].shape) == 1:
-            try:  # TODO: fix this
-                unique_vals = np.unique(dynamic_table[name].data)
-                if 1 < len(unique_vals) <= (len(dynamic_table[name].data) / 2):
+            try:
+                if isinstance(dynamic_table[name].data[0], (str, int, float)):
+                    column_data = [dynamic_table[name].data[i] for i in region]
+                elif hasattr(dynamic_table[name].data[0], 'name'):
```
Saksham20, line 27: This is particularly useful for example if we want column name 'group' containing `ElectrodeGroup`s to be a category. We can get `ElectrodeGroup.name` to categorize traces (this is a possibility in both `Units `and `Electrodes `table which contain `ElectrodeGroup` in a column)

nwbwidgets/utils/dynamictable.py
```diff
@@ -1,21 +1,41 @@
...
+    """
     categorical_cols = {}
+    region = region if region is not None else list(range(len(dynamic_table)))
     for name in dynamic_table.colnames:
         if len(dynamic_table[name].shape) == 1:
-            try:  # TODO: fix this
-                unique_vals = np.unique(dynamic_table[name].data)
-                if 1 < len(unique_vals) <= (len(dynamic_table[name].data) / 2):
+            try:
+                if isinstance(dynamic_table[name].data[0], (str, numbers.Number)):
```
bendichter, line 25: should also include `bytes`

nwbwidgets/utils/dynamictable.py
```diff
@@ -1,21 +1,41 @@
...
+                    column_data = [dynamic_table[name].data[i] for i in region]
+                elif hasattr(dynamic_table[name].data[0], 'name'):
+                    column_data = [dynamic_table[name].data[i].name for i in region]
+                else:
+                    column_data = []
```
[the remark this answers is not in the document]
Saksham20: when an empty list, line 32 will always be false and the `categorical_cols` will not be updated anyway

nwbwidgets/utils/dynamictable.py
Saksham20: oh yes correct. fixed here https://github.com/NeurodataWithoutBorders/nwb-jupyter-widgets/pull/157/commits/558b4077f29ea9ed593a3a4c0fb3f755328ad598

nwbwidgets/utils/dynamictable.py
bendichter: sure, this makes sense

nwbwidgets/controllers/group_and_sort_controllers.py
```diff
@@ -281,8 +285,8 @@ def get_group_vals(self, by, units_select=()):
         """
         if by is None:
             return None
-        elif by in self.dynamic_table:
-            return self.dynamic_table[by][:][units_select]
+        elif by in self.groups:
+            return self.groups[by] if units_select is None else self.groups[by][units_select]
```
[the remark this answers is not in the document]
Saksham20: `self.groups[by]` returns a numpy array, and indexing with `units_select=None` adds a dimension to the array. is there a better way is to index with a default declared value ? One way could be to set `units_select `to `range(len(self.groups[by])`.

Initially, it was pulling column data directly from the dynamic table thus `dynamic_table[by]`  was a hdf dataset so the default value of [()] retrieved all contents. Before, using the `dynamic_table` was essential as there was no other way to pull complete column data from the colname. However, now the output of the `infer_categorical_columns()` contains this data so I'm directly pulling it from the `self.groups` which is set as `self.groups=infer_categorical_columns()` in `__init__`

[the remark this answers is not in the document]
bendichter: OK, makes sense. Can you also use it here?

https://github.com/NeurodataWithoutBorders/nwb-jupyter-widgets/blob/dcb0c7df214c4040f9d34aede6aa78ab751d82cc/nwbwidgets/controllers/group_and_sort_controllers.py#L160-L168

bendichter:
@Saksham20 this look good to me! Is it ready to go from your end?

PR #158, opened 2021-05-14
fix display of unit ids in spiking unit widgets

CodyCBakerPhD:
@bendichter Is there someone I should ping as a point-person for the widgets?

This PR fixes the display of unit ids in session raster widget as well as dropdowns in PSTH and RasterGrid widgets - before, they displayed unit ids as the index to the units table, not the unit id itself. Now appears as
image

### DIFF
--- a/nwbwidgets/misc.py
+++ b/nwbwidgets/misc.py
@@ -106,6 +106,7 @@ def show_session_raster(
         progress_bar=progress_bar,
     )
     ax.set_ylabel("unit #")
+    ax.set_yticklabels(units.id.data[:])
 
     return ax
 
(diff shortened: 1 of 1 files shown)


nwbwidgets/misc.py
```diff
@@ -286,7 +287,7 @@ def __init__(
         if unit_controller is None:
             nunits = len(units["spike_times"].data)
             self.unit_controller = widgets.Dropdown(
-                options=[x for x in range(nunits)],
+                options=[(str(units.id.data[x]), x) for x in range(nunits)],
```
bendichter, line 290: This looks great. The only problem is that his is an unnecessarily high number of H5 Dataset read, which will be slow if there are many units. It would be better to create an ids `np.array` first:
```suggestion
                ids = units.id.data[:]
                options=[(str(ids[x]), x) for x in range(nunits)],
```

nwbwidgets/misc.py
```diff
@@ -836,7 +837,7 @@ def __init__(
 
         trial_event_controller = make_trial_event_controller(self.trials)
         unit_controller = widgets.Dropdown(
-            options=range(len(units["spike_times"].data)),
+            options=[(str(units.id.data[x]), x) for x in range(len(units["spike_times"].data))],
```
bendichter, line 840: same here

bendichter:
@CodyCBakerPhD I can review these. Minor, change, otherwise looks good!

PR #159, opened 2021-05-24
More unit id fixes for unit plots

CodyCBakerPhD:
@bendichter A couple of those fixes we talked about last time

Everything appears visually the same as [PR #158](https://github.com/NeurodataWithoutBorders/nwb-jupyter-widgets/pull/158), just in all the appropriate situations now.


nwbwidgets/misc.py
```diff
@@ -106,7 +106,11 @@ def show_session_raster(
         progress_bar=progress_bar,
     )
     ax.set_ylabel("unit #")
-    ax.set_yticklabels(units.id.data[:])
+    if len(data) <= 30:
+        unit_id_display = np.array(units.id.data[:])[[x for x in this_iter]]
+        ax.set_yticklabels(unit_id_display)
+    else:
+        ax.axes.yaxis.set_visible(False)
```
CodyCBakerPhD, line 113: There's no good way to view the unit ids if they aren't consecutive and more than 30 in number, so just removing the labels altogether is best imo. Could add a test to see if they are at least consecutive if you want

nwbwidgets/misc.py
```diff
@@ -106,7 +106,11 @@ def show_session_raster(
         progress_bar=progress_bar,
     )
     ax.set_ylabel("unit #")
-    ax.set_yticklabels(units.id.data[:])
+    if len(data) <= 30:
+        unit_id_display = np.array(units.id.data[:])[[x for x in this_iter]]
```
CodyCBakerPhD, line 110: The previous version did not actually filter the unit ids based on the unit selector bar - this version does.

nwbwidgets/misc.py
```diff
@@ -285,10 +289,10 @@ def __init__(
             self.trials = trials
 
         if unit_controller is None:
-            unit_ids = units.id.data[:]
-            n_units = len(unit_ids)
+            self.unit_ids = units.id.data[:]
```
CodyCBakerPhD, line 292: The unit ids are  used across more than one function in this class, so I just set it as an attribute for a single IO operation

PR #160, opened 2021-05-30
Group sort controller updates

Saksham20:
integration of range_controller within the group and sort controller. 

If there are no categorical columns found for the `DynamicTable`, then the output of `GroupAndSortController `is a `RangeController`.

This works on top of PR #152 . Changes are only in the group_and_sort_controllers.py and timeseries.py. I think the diff will be easier to review after the merge of #152


Saksham20:
@bendichter tests now pass for the given changes, lets review and discuss this after the PR #152 merge

nwbwidgets/controllers/group_and_sort_controllers.py
```diff
@@ -39,10 +38,10 @@ def __init__(self, dynamic_table: DynamicTable):
 class GroupAndSortController(AbstractGroupAndSortController):
     def __init__(
         self,
-        dynamic_table: DynamicTable,
+        dynamic_table: DynamicTable=None,
```
bendichter, line 41: I do not understand this change. How can a GroupAndSortController work without a DynamicTable?

Saksham20: if there is no dynamic table present, then essentially this controller will revert to a range_controller for the y axis. This requires the keep_rows (if dt is None) to get the correct range). If keep rows is given with dynamic_table then it applies that to this table instead as per the old design. A better name for 'GroupAndSort' is still pending though

bendichter: I'd rather not do it this way. If a user wants a range controller and does not have a dynamic table to input, they should just use RangeController. If we need a range controller with keep_rows, we should add it to RangeController instead with an arg name like "all_values"

Saksham20: I think its lesser on the Modular scale but a little more elegant and saves some lines of code on Classes implementing this controller. One possibility is to rename this to something like `RowsRangeController `and have an arg as `range_controller=False` and `range: list=None` argument (will be provided only if `range_controller=True`) to make things more explicit. 
I feel the group and sorting functionality is a valid subset of `RowsRangeController` (or a similar name) since grouping and sorting process decides the valid range of the rows (if a dynamictable is provided that is) and if not then select all rows.  So I think having it under one class could be feasible at the cost of modularity though. What do you think?

bendichter: what do you think? Would that work? Would it accomplish what you are trying to accomplish here?

bendichter: I don't understand what you mean. I think if a user does not have a DynamicTable and simply wants to control a range, they should use RangeController. If they do have a DynamicTable and want to do grouping, they should use GroupAndSortController.

Saksham20:
i think the merge master reverted many changes that I had made.. was it done automatically by github?

Saksham20:
please dont merge it to master.. once approved i'll merge it offline and push

bendichter:
ok

nwbwidgets/controllers/group_and_sort_controllers.py
```diff
@@ -293,13 +307,12 @@ def get_orderable_cols(self):
         return [x for x in candidate_cols if len(robust_unique(self.units[x][:])) > 1]
 
     def group_and_sort(self):
-        if self.group_vals is None and self.order_vals is None:
+        if self.column_values is None and self.order_vals is None:
```
bendichter, line 310: why are you changing the name of this variable from group_vals to column_values?

nwbwidgets/controllers/group_and_sort_controllers.py
```diff
@@ -258,7 +276,7 @@ def range_controller_observer(self, change):
     def get_groups(self):
         return infer_categorical_columns(self.dynamic_table, self.keep_rows)
 
-    def get_group_vals(self, by, units_select=None):
+    def get_column_values(self, by, units_select=None):
```
bendichter, line 279: For consistency, can you change unit_select to keep_rows?

nwbwidgets/utils/dynamictable.py
```diff
@@ -18,7 +18,7 @@ def infer_categorical_columns(dynamic_table: DynamicTable, region: Iterable = No
         keys: as columns that are categorical, values as the unique values
     """
     categorical_cols = {}
-    region = region if region is not None else list(range(len(dynamic_table)))
+    region = region if region is not None else np.arange(len(dynamic_table))
```
bendichter, line 21: can you use dtype='uint' here?

bendichter:
@Saksham20 this is looking really good! Sorry about messing up the PR earlier. Just two minor changes, then go ahead and merge it when you have time

Saksham20:
> @Saksham20 this is looking really good! Sorry about messing up the PR earlier. Just two minor changes, then go ahead and merge it when you have time

@bendichter you need to approve the review so that it allows me to manually merge to master

bendichter (approved):

PR #162, opened 2021-06-09
moving matplotlib to plotly

Saksham20:
moving all instances of matplotlib plots to plotly figureWidgets


bendichter:
Thanks, this is looking good! Can you please include screenshots of each converted figure?

bendichter:
This PR is getting very big. Would it be possible to break it into smaller PRs, each with proper screenshot and tests?

PR #164, opened 2021-06-11
window control for spatial traces

### LINKED ISSUE #163: use time window controller for X by Y view of SpatialSeries
Right now the "trace" tab of a SpatialSeries object shows this view:

image

All timepoints are plotted, which often takes a long time and results in unintelligible position traces. It would be better to replace this with a widget that has a time range controller.

bendichter:
fix #163 

image

### DIFF
--- a/nwbwidgets/behavior.py
+++ b/nwbwidgets/behavior.py
@@ -99,11 +99,11 @@ def route_spatial_series(spatial_series, **kwargs):
 
 class SpatialSeriesTraceWidget(AbstractTraceWidget):
     @abstractmethod
-    def plot_data(self, data, units):
+    def plot_data(self, data, units, tt):
         return
 
     @abstractmethod
-    def update_plot(self, data):
+    def update_plot(self, data, tt):
         return
 
     def __init__(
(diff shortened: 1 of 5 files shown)


oruebel:
How about coloring the trace by time and adding a "Play" button to animate the plot.

bendichter:
@oruebel good ideas. Let me see how straightforward that would be in plotly. I like the idea of the play button but it might be more than I am able to work on right now.

oruebel:
>  I like the idea of the play button but it might be more than I am able to work on right now.

No worries. Just some random ideas.

bendichter:
I was able to get the color working. Let's move the play button feature to another PR. That would ideally be a dedicated controller that we can reuse for other widgets.

oruebel:
>  Let's move the play button feature to another PR

That makes sense. For Plotly a basic tutorial for animations is here https://plotly.com/python/animations/ and there is also a play widget https://ipywidgets.readthedocs.io/en/latest/examples/Widget%20List.html#Play-(Animation)-widget . I think for Plotly you may need all the data loaded whereas for the play widget I think you are just attaching a python callback function to the buttons as an observer so with that you should be able load data on demand and plot. I think it'll depend somewhat on the size of the data whether you want to load data at once (and then have smooth animation) or load data on demand.

nwbwidgets/behavior.py
```diff
@@ -95,47 +72,134 @@ def show_spatial_series(node: SpatialSeries, **kwargs):
...
+    def plot_data(self, data, units, tt):
+        if units is None:
+            units = "no units"
+
+        self.out_fig = go.FigureWidget(
+            data=go.Scatter3d(
+                x=data[:, 0],
+                y=data[:, 1],
+                z=data[:, 2],
+                # mode="lines",
```
Saksham20, line 182:
```suggestion
                mode="markers",
```
Being explicit about the mode makes the color scale appear on my browser. 
I think this is a great way to visually show 1) the speed of movement 2) places there can be discontinuity due to missing inter-trial data. Super!

nwbwidgets/behavior.py
```diff
@@ -95,47 +72,134 @@ def show_spatial_series(node: SpatialSeries, **kwargs):
...
+
+class SpatialSeriesTraceWidget2D(SpatialSeriesTraceWidget):
+    def plot_data(self, data, units, tt):
+        if units is None:
+            units = "no units"
+        self.out_fig = go.FigureWidget(
+            data=go.Scatter(
+                x=list(data[:, 0]),
+                y=list(data[:, 1]),
+                # mode="lines",
```
Saksham20, line 152:
```suggestion
                mode="markers",
```

Saksham20 (approved):

PR #168, opened 2021-06-14
updating plotly function to show timeseries

Saksham20:
update `show_index_timeseries_plotly()` and its usage in `SingleTracePlotlyWidget `and `SeparateTracesPlotlyWidget`

### DIFF
--- a/nwbwidgets/test/test_timeseries.py
+++ b/nwbwidgets/test/test_timeseries.py
@@ -1,5 +1,5 @@
 import unittest
-
+import plotly.graph_objects as go
 import matplotlib.pyplot as plt
 import numpy as np
 from ipywidgets import widgets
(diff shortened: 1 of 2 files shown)


Saksham20:
correctly showing behavior time series with this new change to `show_index_timeseries_plotly()`
separatetraces

bendichter:
@Saksham20 This looks good! Nice job with the tests and the gif. One thing: sometimes it is useful to be able to input the index instead of the time. Could you change this so that it is able to take as input either indices or times (but not both)?

bendichter:
Then could you add tests for inputting index and a test making sure that inputting both raises a value error

bendichter (changes requested):
see above

bendichter (approved):

PR #169, opened 2021-06-15
plotly for AlignMultiTraceTimeSeriesByTrials

Saksham20:
- [x] using plotly for the behavior classes: `AlignMultiTraceTimeSeriesByTrialsVariable` and `AlignMultiTraceTimeSeriesByTrialsConstant`. 
- [x] tests
- [x] gif
- [x] figure slow to load issue.

### DIFF
--- a/nwbwidgets/controllers/group_and_sort_controllers.py
+++ b/nwbwidgets/controllers/group_and_sort_controllers.py
@@ -203,6 +203,7 @@ class GroupAndSortController(AbstractGroupAndSortController):
             self.group_sm.options = tuple(groups[::-1])
             self.group_sm.value = self.group_sm.options[:1]
             self.group_sm.disabled = False
+            self.selected_column_values = self.group_sm.value
             if self.control_limit:
                 self.limit_bit.disabled = False
                 self.limit_cb.disabled = False
--- a/nwbwidgets/controllers/group_and_sort_controllers.py
+++ b/nwbwidgets/controllers/group_and_sort_controllers.py
@@ -201,7 +201,7 @@ class GroupAndSortController(AbstractGroupAndSortController):
             groups = np.unique(keep_column_values)
             self.group_sm.rows = min(len(groups), 20)
             self.group_sm.options = tuple(groups[::-1])
-            self.group_sm.value = self.group_sm.options
+            self.group_sm.value = self.group_sm.options[:1]
             self.group_sm.disabled = False
             if self.control_limit:
                 self.limit_bit.disabled = False
--- a/nwbwidgets/controllers/group_and_sort_controllers.py
+++ b/nwbwidgets/controllers/group_and_sort_controllers.py
@@ -20,15 +20,16 @@ class AbstractGroupAndSortController(widgets.VBox, ValueWidget):
     )
     """
 
-    def __init__(self, dynamic_table: DynamicTable):
+    def __init__(self, dynamic_table: DynamicTable, keep_rows=None):
         super().__init__()
-
+        self.keep_rows = (
+            keep_rows if keep_rows is not None else np.arange(len(dynamic_table))
+        )
         self.dynamic_table = dynamic_table
-        self.nitems = len(self.dynamic_table.id)
-
-        self.group_vals = None
+        self.nitems = len(self.keep_rows)
+        self.column_values = None
         self.group_by = None
-        self.group_select = None
+        self.selected_column_values = None
         self.limit = None
         self.desc = False
         self.order_by = None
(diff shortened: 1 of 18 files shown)


nwbwidgets/controllers/group_and_sort_controllers.py
```diff
@@ -201,7 +201,7 @@ def set_group_by(self, group_by):
             groups = np.unique(keep_column_values)
             self.group_sm.rows = min(len(groups), 20)
             self.group_sm.options = tuple(groups[::-1])
-            self.group_sm.value = self.group_sm.options
+            self.group_sm.value = self.group_sm.options[:1]
```
bendichter, line 204: What is this doing?

Saksham20: In my attempts to investigate why plotting is slow when I change the values of group and sort, I came across one of these possibilities: after selecting a col by which to group by, the  select multiple drop down widget defaults to all group values at the start, which then cascades down to the plot adding all of those traces. So I think it might be better to default to one value from the drop down and the user can explicitly select multiple if needed. Check the image below (it defaults to 12 unlike before where it default to all values): 
image

bendichter: Makes sense. We could even select the empty set to start

Saksham20: yeah thats even better actually! changed it https://github.com/NeurodataWithoutBorders/nwb-jupyter-widgets/pull/169/commits/0e21776323d42b95b7919a37caee4b773c1203bb

Saksham20:
alignmultitracetimeseries

bendichter:
I was thinking about this and I wonder if the speed issue might be due to 
1. updating the plot 1000 times, fixed with `with fig.batch_update()`
2. slicing the `TimeSeries.data` h5py dataset 1000 times. You can fix this by pre-reading a region of the dataset and slicing the numpy array instead. 

@Saksham20, Can you try these and see if they fix the speed issue?

bendichter:
I don't think memoization is going to help us much here

Saksham20:
> I don't think memoization is going to help us much here

it helps reduce the time to loading the aligned data to almost 0 sec from 15 secs

bendichter:
@Saksham20 

import plotly.graph_objects as go

data1 = [5,4,3,4,5]
data2 = [7,6,5,6,7]
data3 = [9,8,7,8,9]
data_list = [data1, data2, data3]
tt = [1,2,3,4,5]

# method 1 (each trial is a unique trace)

fig = go.FigureWidget()
for data in data_list:
    fig.add_trace(go.Scattergl(x=tt, y=data, marker_color="blue"))
fig

# method 2 (all of the data is a single trace)

import numpy as np
import copy

def inject_nans(data_list):
    data_list = copy.copy(data_list)
    for x in range(len(data_list)-1, 0, -1):
        data_list.insert(x, np.nan)
    return np.hstack(data_list)

all_data = inject_nans(data_list)
all_tt = inject_nans([tt]*len(data_list))

fig = go.FigureWidget()
fig.add_trace(go.Scattergl(x=all_tt, y=all_data))
fig

Saksham20:
> @Saksham20
> 
> ```python
> import plotly.graph_objects as go
> 
> data1 = [5,4,3,4,5]
> data2 = [7,6,5,6,7]
> data3 = [9,8,7,8,9]
> data_list = [data1, data2, data3]
> tt = [1,2,3,4,5]
> 
> # method 1 (each trial is a unique trace)
> 
> fig = go.FigureWidget()
> for data in data_list:
>     fig.add_trace(go.Scattergl(x=tt, y=data, marker_color="blue"))
> fig
> 
> # method 2 (all of the data is a single trace)
> 
> import numpy as np
> import copy
> 
> def inject_nans(data_list):
>     data_list = copy.copy(data_list)
>     for x in range(len(data_list)-1, 0, -1):
>         data_list.insert(x, np.nan)
>     return np.hstack(data_list)
> 
> all_data = inject_nans(data_list)
> all_tt = inject_nans([tt]*len(data_list))
> 
> fig = go.FigureWidget()
> fig.add_trace(go.Scattergl(x=all_tt, y=all_data))
> fig
> ```

thanks @bendichter ,insertion of nans to make multiple traces as one really makes a difference in plotting times. I think the bottleneck was the rendering time.

bendichter:
@Saksham20 this is very close! Let's move the trace merging trick to the `multi_trace` function in `utils/plotly.py`

Saksham20:
> @Saksham20 this is very close! Let's move the trace merging trick to the `multi_trace` function in `utils/plotly.py`

done [here](https://github.com/NeurodataWithoutBorders/nwb-jupyter-widgets/pull/169/commits/dc799838cc6e047076a6921de16d761df4940860)

bendichter (approved):

PR #180, opened 2021-11-09
Added DynamicSummaryTable

alejoe91:
The `DynamicSummaryTable` allows to inspect different dynamic tables. It allows to select multiple columns and it adjusts the rendering accordingly:

- 1 real column: histogram + marginal violin
- 1 cat column: bar plot
- 2 real columns: scatter + marginal violins
- 1 real + 1 cat columns: multiple histograms + marginal violins
- 2 real + 1 cat columns: 2d scatter + color (for category)
- 3 real variable: 3d scatter
- 3 real + 1 cat columns: 3d scatter + color (for category)

Here is a quick demo:

https://user-images.githubusercontent.com/17097257/140964089-f5c3b999-d94f-4cb9-ae01-c79963b24c2e.mov

- [x] add tests


bendichter:
what happens if the user selects more than 3 columns?

bendichter:
non-critical: for categorical variables I would rather a legend or a colorbar with discrete colors rather that a continuously varying colorbar

bendichter:
(I love this by the way!)

bendichter:
We could also add 3 real + 1 cat: 3d scatter + color

alejoe91:
> what happens if the user selects more than 3 columns?

Only three columns are forced by resetting the value of the `SelectMultiple` widget

nwbwidgets/dynamictablesummary.py
```diff
@@ -0,0 +1,144 @@
...
+
+import matplotlib.pyplot as plt
+
+
+field_lay = widgets.Layout(
+        max_height="40px", max_width="500px", min_height="30px", min_width="180px"
+    )
+
+
+class DynamicTableSummary(widgets.VBox):
```
bendichter, line 21: Can we call this `DynamicTableSummaryWidget`?

nwbwidgets/dynamictablesummary.py
```diff
@@ -0,0 +1,144 @@
...
+
+from .utils.dynamictable import infer_categorical_columns, infer_columns_to_plot, infer_columns_to_plot
+
+from .utils.widgets import interactive_output
+
+import plotly.graph_objects as go
+
+import plotly.express as px
+
+import matplotlib.pyplot as plt
```
bendichter, line 13:
```suggestion
from ipywidgets import widgets, fixed, FloatProgress, Layout
import matplotlib.pyplot as plt
import plotly.graph_objects as go
import plotly.express as px
from pynwb.misc import AnnotationSeries, Units, DecompositionSeries, DynamicTable

from .utils.dynamictable import infer_categorical_columns, infer_columns_to_plot, infer_columns_to_plot
from .utils.widgets import interactive_output
```

test/test_dynamictablesummary.py
```diff
@@ -0,0 +1,114 @@
...
+        cols_to_plot = (real_names[0], real_names[1], real_names[2])
+        widget.controls["col_names_display"] = cols_to_plot
+        widget.plot_hist_bar(col_names_display=cols_to_plot, nbins=nbins)
+        
+
+        # select 3 real + 1 cat
+        cols_to_plot = (real_names[0], real_names[1],
+                        real_names[2], cat_names[0])
+        widget.controls["col_names_display"] = cols_to_plot
+        widget.plot_hist_bar(col_names_display=cols_to_plot, nbins=nbins)
```
bendichter, line 114: @alejoe91 would it be possible to change the dropdown and have the plot work automatically, like this?

```suggestion
        widget.controls.column_dropdown.value = cols_to_plot
```

bendichter: Could you also add lines for each plot type, and that tests the error catching? I'd like the new widget to be fully covered by tests.

alejoe91: No exceptions are thrown, but the dropdown is reset if combinations are not supported. Adding tests for those now

alejoe91:
@bendichter I added tets that trigger a reset of the dropdown and suggest possible combinations https://github.com/catalystneuro/nwb-jupyter-widgets/blob/dynamictablesummary/test/test_dynamictablesummary.py#L138-L155)

I also added the possibility to show labels on the graph, which will be useful to understand which entry is which. See attached video:

https://user-images.githubusercontent.com/17097257/141811371-94afa68b-d8a8-4568-ac35-fbc4012b1274.mov

bendichter (approved):

bendichter:
Good to merge once tests pass

PR #196, opened 2022-01-05
Two photon series fixes

Saksham20:
Updating the `ImageSeriesWidget` and making `TwoPhotonSeriesWidget` as a child class. 

1. Changes to `ImageSeriesWidget`:
- in case of `external_file`:
   - Formats supported: Tif > [Tif, mp4, avi.. videos] 
   - Creates an additional widget which helps the user choose which of the file from the list of `external_file` needs to be viz.  

2. Changes to `TwoPhotonSeriesWidget`: 
- Inherits from `ImageSeriesWidget`

### DIFF
--- a/nwbwidgets/image.py
+++ b/nwbwidgets/image.py
@@ -43,10 +43,9 @@ class ImageSeriesWidget(widgets.VBox):
     ):
         super().__init__()
         self.imageseries = imageseries
-        self.controls = {}
-
+        self.figure = None
+        self.time_window_controller = foreign_time_window_controller
         tmin = get_timeseries_mint(imageseries)
-
         # Make widget figure --------
         def _add_fig_trace(img_fig: go.Figure, index):
             if self.figure is None:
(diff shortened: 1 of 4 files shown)


Saksham20:
widgets_ImageSeries_demo

nwbwidgets/image.py
```diff
@@ -1,94 +1,159 @@
...
+                self.figure.for_each_trace(lambda trace: trace.update(img_fig.data[0]))
+            self.figure.layout.title = f"Frame no: {index}"
+
+        def _add_time_slider_controller(min, max):
+            if self.time_slider is None:
+                self.time_slider = widgets.FloatSlider(value=0,
+                                                       min=min,
+                                                       max=max,
+                                                       orientation="horizontal",
+                                                       description="time(s)")
```
bendichter, line 64: Why are you defining these functions inside the __init__?

nwbwidgets/image.py
```diff
@@ -1,94 +1,159 @@
...
             else:
-                tmax = get_timeseries_maxt(imageseries)
-            self.time_window_controller = StartAndDurationController(tmax, tmin)
+                self.time_slider.max = max
+                self.time_slider.min = min
+                self.time_slider.value = min
+
+        if imageseries.external_file is not None:
+            file_selector = widgets.Dropdown(options=imageseries.external_file)
```
bendichter, line 71: there should only be a dropdown if there is more than one external file

nwbwidgets/image.py
```diff
@@ -167,3 +232,54 @@ def show_rbga_image(rgb_image: RGBImage, neurodata_vis_spec=None):
...
+        return frame.shape
+
+
+def get_frame_count(external_path_file: PathType):
+    external_path_file = Path(external_path_file)
+    if external_path_file.suffix in ['.tif', '.tiff']:
+        assert HAVE_TIF, 'pip install tifffile'
+        tif = TiffFile(external_path_file)
+        return len(tif.pages)
+    else:
```
bendichter, line 258: should be elif, and else should be an error message saying this type of external file is not yet supported by nwb widgets

nwbwidgets/image.py
```diff
@@ -167,3 +232,54 @@ def show_rbga_image(rgb_image: RGBImage, neurodata_vis_spec=None):
...
+        assert HAVE_OPENCV, 'pip install opencv-python'
+        cap = cv2.VideoCapture(str(external_path_file))
```
bendichter, line 260: could we use a context manager for this? https://stackoverflow.com/questions/19086343/with-statement-to-read-from-videocapture-in-opencv

nwbwidgets/image.py
```diff
@@ -1,94 +1,159 @@
...
+            return int((time - self.imageseries.starting_time)*self.imageseries.rate)
         else:
             return timeseries_time_to_ind(self.imageseries, time)
 
-    def set_controls(self, **kwargs):
-        self.controls.update(
-            timeseries=fixed(self.imageseries), time_window=self.time_window_controller
-        )
-        self.controls.update({key: widgets.fixed(val) for key, val in kwargs.items()})
+    def set_children(self, *args):
```
bendichter, line 147: why the *args?

nwbwidgets/ophys.py
```diff
@@ -14,89 +13,23 @@
...
-                max=indexed_timeseries.data.shape[0] - 1,
-                orientation="horizontal",
-            )
-
-        slider.observe(lambda change: update_figure(change.new), names="value")
-        self.figure = None
-        self.controls = dict(slider=slider)
-        update_figure()
-        self.children = [self.figure, slider]
+        super(TwoPhotonSeriesWidget, self).__init__(indexed_timeseries,
```
bendichter, line 31:
```suggestion
        super().__init__(indexed_timeseries,
```

Saksham20:
TODO:

- [ ] context manager for videos
- [ ] create self.figure in set_figure
- [ ] move nested functions to class methods
- [ ] try to create the image trace instead of using plotly express
- [ ] refactor file-format-specific functions into separate functions
- [ ] use black
- [ ] add back foreign time window controller
- [ ] add tests

nwbwidgets/image.py
```diff
@@ -1,94 +1,173 @@
...
+                + get_frame_count(imageseries.external_file[0]) / imageseries.rate
+            )
+            if self.time_slider is None:
+                self.time_slider = widgets.FloatSlider(
+                    min=imageseries.starting_time,
+                    max=tmax,
+                    orientation="horizontal",
+                    description="time(s)",
+                )
+            external_file = imageseries.external_file[0]
```
bendichter, line 51: I think it would be easier if you set this as a class attribute

nwbwidgets/image.py
```diff
@@ -1,94 +1,173 @@
...
+            self.file_selector = None
+            # set file selector:
+            if len(imageseries.external_file) > 1:
+                self.file_selector = widgets.Dropdown(options=imageseries.external_file)
+                external_file = self.file_selector.value
+
+                def update_time_slider(value):
```
bendichter, line 58: I think it would be easier if this were a method

nwbwidgets/image.py
```diff
@@ -1,94 +1,173 @@
...
+                    path_ext_file = value["new"]
+                    # Read first frame
+                    nonlocal external_file
+                    external_file = path_ext_file
+                    tmax = (
+                        imageseries.starting_time
+                        + get_frame_count(path_ext_file) / imageseries.rate
+                    )
+                    tmin = 0
```
bendichter, line 67: why reset tmin to 0 every update?

nwbwidgets/image.py
Saksham20: all callbacks are now methods:  https://github.com/NeurodataWithoutBorders/nwb-jupyter-widgets/pull/196/commits/8e72ae96033c16009e718c501f7c62666f6ed2c8

nwbwidgets/image.py
Saksham20: Yes with callbacks now set as methods, it makes sense to have it as an attribute: https://github.com/NeurodataWithoutBorders/nwb-jupyter-widgets/commit/8e72ae96033c16009e718c501f7c62666f6ed2c8

nwbwidgets/image.py
```diff
@@ -1,94 +1,167 @@
...
             return timeseries_time_to_ind(self.imageseries, time)
 
-    def set_controls(self, **kwargs):
-        self.controls.update(
-            timeseries=fixed(self.imageseries), time_window=self.time_window_controller
-        )
-        self.controls.update({key: widgets.fixed(val) for key, val in kwargs.items()})
+    def get_children(self, *widgets):
+        set_widgets = [wid for wid in widgets if wid is not None]
+        return [self.figure, self.time_slider, *set_widgets]
```
bendichter, line 158: a foreign time slider should not be included in the children

nwbwidgets/image.py
Saksham20: knowing the `starting_time` for video files other than the first one is not possible. Unless we assume the starting time to the end time for the last video in the sequence. But I don't think this would be a good approximation, so I have made  the `time_slider` widget to span the video frames (0, num_frames) instead, only when the user requests video other than the first one.

nwbwidgets/image.py
Saksham20: In case there is a `foreign_time_slider`, then what is the best way to add another callback (update figure) ? The usual way of using `foreign_time_slider.observe(set_figure)` would overwrite the old callbacks. 

One possible solution would be to link the values of `foreign_time_slider` to the new `time_slider`: `widgets.jslink((foreign_time_slider,"value"),(time_slider,"value"))`  In effect, the time_slider would now be controlled by the foreign slider while keeping the functionality of being changed independently. 

Check https://github.com/NeurodataWithoutBorders/nwb-jupyter-widgets/pull/196/commits/ae8e9cc42181bc6a21ea2d7786180cfd27b4eb7c

PR #203, opened 2022-02-22
fix gain for timeseries

luiztauffer:


luiztauffer:
@bendichter a neo reader from ABF files yields two types of scale values:
- a dimension scale (V, mV, uV)
- a scale from int16 to the dimension above

In the nwbfile, I’ve been storing these values as below:
image

luiztauffer:
@bendichter I believe the gain should be applied to the data from these files, the traces seems to be around the expected voltage when done so:
image

CodyCBakerPhD:
@luiztauffer All NWB TimeSeries expect the `conversion` factor to scale the raw data storage into the specified units, in this case volts. In your case just incorporate the `gain` from `neo` into the factor.

There is a very special case with `ElectricalSeries` in which you could technically hardness the `n_channel x 1` vector attribute `channel_conversion` in addition to the scalar `conversion` (`data_in_volts = data_in_storage * channel_conversion * conversion` in that case. But I really wouldn't recommend it unless different channels have different gains.

FYI: we're also close to finishing another general attribute to the equation, `data_in_volts = data_in_storage * conversion + offset`.

bendichter:
@luiztauffer I think you should be using the `conversion` field instead. See the notes for gain [here](https://nwb-schema.readthedocs.io/en/latest/format.html?highlight=gain#patchclampseries)

luiztauffer:
@CodyCBakerPhD @bendichter thanks for clarifying 
> In your case just incorporate the gain from neo into the factor.

That is reasonable and easy to do, I'll fix the nwb files and this PR here will be unnecessary then, I'll close it

PR #207, opened 2022-05-12
Input form

luiztauffer:
- Input form to load local nwb files or stream from dandi
- simplified usage with minimal code

Peek 2022-05-12 18-04

### DIFF
--- a/nwbwidgets/dashboard.py
+++ b/nwbwidgets/dashboard.py
@@ -11,7 +11,7 @@ def dashboard():
     source_options_label = widgets.Label('Source:')
     source_options = widgets.VBox([source_options_label, source_options_radio], layout=widgets.Layout(width='150px', overflow=None))
 
-    source_path_text = widgets.Text(value="000239", layout={'width': '300px'})
+    source_path_text = widgets.Text(value="", layout={'width': '300px'})
     source_path_dandi_button = widgets.Button(icon='angle-right', layout={'width': '50px'})
     source_path_lower = widgets.HBox([source_path_text, source_path_dandi_button])
     source_path_label = widgets.Label('Dandiset number:')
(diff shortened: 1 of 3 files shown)


bendichter (approved):

CodyCBakerPhD:
Hey @bendichter, @h-mayorquin was wanting to view some sample files on DANDI to get a feel for the widgets - I was going to instruct him on how to run the widgets manually and resolve S3 paths, etc, but this PR looks so good/ready, is there anything else to merge?

h-mayorquin:
Trying this gives me the known error (after a pip install in the environment):

```
ValueError: h5py was built without ROS3 support, can't use ros3 driver
```

I tried to track down the dependencies but could not resolve if installing dandi or any other package would resolve this (build h5py with that option I guess).

luiztauffer:
@h-mayorquin can you try this solution in a clean environment?
https://github.com/NeurodataWithoutBorders/pynwb/issues/1476

nwbwidgets/dashboard.py
```diff
@@ -0,0 +1,89 @@
+import ipywidgets as widgets
+from pathlib import Path
+
+from pynwb import NWBHDF5IO
+from nwbwidgets import nwb2widget
+from dandi.dandiapi import DandiAPIClient
+
+
+def dashboard():
```
bendichter, line 9: can we call this something else? Like a file selector?

requirements-dev.txt
```diff
@@ -0,0 +1,12 @@
+pynwb
```
bendichter, line 1: why have a different requirements-dev file here? The constraints on ipywidgets is important

nwbwidgets/dashboard.py
luiztauffer: just renamed to `input_form`, hows that? @bendichter

requirements-dev.txt
luiztauffer: this is old, I don't remember why... but we can safely remove it

Issue #209, opened 2022-07-27
Idea: make some dependencies optional


TomDonoghue:
Adding a quick issue here to note / log a brief convo @ NWB user days about maybe supporting optional dependencies.

The pros here are the potential to make the module more light-weight, the cons are that it complicated installation in at least some cases in which users have to now / figure out something about adding optional dependencies they need.

Practically, the `extras_require` in `setup.py` can handle installing different sets of requirements, though I don't see this used all that much, so people might not be all that familiar with it. 

I've added optional dependencies to modules before ([example](https://github.com/spiketools/spiketools/blob/main/spiketools/modutils/dependencies.py)), and in general I think this can be technically handled fairly easily, the main ToDo here is more to figure out a useful architecture for organizing which dependencies are more / less critical, and how the widget should proceed if / when it encounters something it needs for a particular data structure.

TomDonoghue:
Related to this, `nwbwidgets.brains` has some unlisted / optional dependencies (scikit-spatial & nilearn) that are imported within related functions, but not currently listed as dependencies or installed, etc. 

This file is not currently tested (such that it doesn't raise any errors there), but in `ecephys` it is listed for plotting electrodes if the species is set as "Homo sapiens", and would (I presume) currently fail inelegantly if a datafile has this kind of data, but isn't installed to include the extra dependencies.

TomDonoghue:
Closing this issue since it has been completed in #219

PR #211, opened 2022-07-27
[MNT] - Sort and group imports

TomDonoghue:
This PR groups and orders imports across the module. 

Commits:
- sort and group import statements
- update imports for consistency (importing whole module vs. `from x`, etc)

The changes are mostly aesthetic here. In a few cases, this also removes duplicated updates, and/or changes to importing and accessing an object directly, rather than from a general import. 

This update is conceptually somewhat related to #209 - with the idea being that grouping and sorting here makes it easier for scanning through and seeing what is used where, and which dependencies could be candidates for making optional.


bendichter:
looks good, thanks @TomDonoghue!

TomDonoghue:
I added an extra commit that does the same import sorting in the test files. 

Sidenote: `pytest` is used in one test file, so this should maybe be considered a dev requirement (relates to #212).

Also: sorry about the auto-formatter that removes the trailing white space, I didn't realize it was doing that. I didn't explicitly change anything below the imports, so any diff in code files should only be trailing whitespace removal.

Issue #217, opened 2022-08-01
[BUG] - Error when viewing a empty TimeSeries slice


TomDonoghue:
In our experiment we have some TimeSeries that only span certain ranges (for example, in a video game task position and speed are only defined navigation periods). When using the slider, it can happen that one selects a time period with no values. If so, an error is raised because the plot update tries to check the data range (for setting y-axis limits) and then can't on an empty slice. 

ToDo: add a check here (timeseries.py, in `on_change` ~ line 348) that checks for an empty array and deals with it nicely. 

Full error log:
<img width="984" alt="nwbwidgets-minrangeerror" src="https://user-images.githubusercontent.com/7727566/182168687-a5fa5eff-7562-41fb-a8dd-4186967ad3c7.png">

Note: this is somewhat related to #67, but more restricted - for now at least, displaying an empty segment is fine, the idea here is just to stop the error raising.

TomDonoghue:
Fixed in #218

PR #218, opened 2022-08-01
[BUG] - Update timeseries range checks

TomDonoghue:
Responds to #217 

The current update updates the `SingleTracePlotyWidget` to deal with displaying a selected segment that happens to be empty. 

Updates:
- add a try / except for setting y-axis limits, and default to None if no data
- update x-axis limits to show the selected range, rather than the x-range from the data
    - Note: this changes behaviour a bit: previously updating data would autoscale to the range of the selected data (so if selecting 30 seconds, but data was only in range 5-25, it would show 5-25), whereas with this update the x-range will always match the selected range. This is an easy way to address this specific issue, but might not be a preferred way to do it? (An alternative is to add an extra try / except, and show empty range if no data, auto-scale if there is _any_ data (?))

Notes:
- if this solution is considered a good way to go, I think it probably wants to be similarly updated in `SeparateTracesPlotlyWidget` (`on_change` - lines ~398-421).

### DIFF
--- a/nwbwidgets/timeseries.py
+++ b/nwbwidgets/timeseries.py
@@ -410,7 +410,8 @@ class SeparateTracesPlotlyWidget(AbstractTraceWidget):
                         self.out_fig.data[i].x = tt
                         self.out_fig.data[i].y = dd
                         self.out_fig.update_yaxes(
-                            range=[min(dd), max(dd)], row=i + 1, col=1
+                            range=[min(dd), max(dd)] if dd.size != 0 else [None, None],
+                            row=i + 1, col=1
                         )
                         self.out_fig.update_xaxes(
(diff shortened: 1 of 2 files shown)


bendichter:
This looks like it would probably work, but I prefer to avoid try/except flows. Would it be possible to implement this by testing of yy is empty instead?

bendichter (approved):

TomDonoghue:
Comments / updates integrated!

Quick notes:
- there was a duplicated section in the timeseries tests, so one commit cleans that up
- at the end I added the same procedure to the multi-timeseries widget, so that the two widgets act the same
- tests are added: i added timestamps to have a gap and then selected an empty time range

PR #219, opened 2022-08-01
[MNT - Add a strategy for managing optional dependencies

TomDonoghue:
Responds to #209 

This is a quick example of a potential way to manage optional dependencies, based on something I've done for this before. 

The basic approach is:
- use a `safe_import` procedure to import modules if available that is used for dependencies considered optional
- when a widget gets called, check if any optional dependencies are available (with a decorator):
    - if yes, proceed as normal, calling the function to return the expected widget
    - if no, returns a Label widget which notes what's missing (this approach means there is no error, and the overall widget accordion otherwise should work as normal)

This is still WIP in that I wanted to check if the strategy makes sense / seems good, before going further on this - if it's a solid option, then some finishing touches would be needed for properly updating which dependencies are listed as optional. 

Here's what it looks like in a test example of a widget that is missing a dependency:
Screen Shot 2022-08-01 at 2 35 53 PM

As an example use case, the "example usage" commit shows how this could be used for some of the specific electrode viewers, in a way that could make {nilearn, skspatial, ccfwidget} be listed as optional dependencies. However - I don't have a full set of files that test these exact widgets - so properly testing these specific cases is still ToDo.


bendichter:
@TomDonoghue this looks perfect! Let me know when it's ready for a thorough review

TomDonoghue:
@bendichter - cool, since this seems like a reasonably strategy, I finished up the rest of this as a proper candidate PR. The updates in the additional commits mostly clean up the use of the optional imports, and apply this across install / tests / documentation. 

The current version of this PR makes the extra requirements for two widgets optional:
- `HumanElectrodesPlotlyWidget`: nilearn, skspatial, trimesh
    - note that `nilearn` and `skspatial` weren't listed as requirements anyways, so this doesn't remove them from being installed
    - the `trimesh` dependency is a bit of a quirk, in that it's currently not actually used (only used in a section of code commented out). For now, I've listed it as an optional dependency for this datatype, but more broadly, this functionality presumably wants to be either turned back on, or maybe be removed (?)
    - this widget is in `brains.py`, which isn't tested (I guess tests should be added for this)
- `show_ccf` (which is like "MouseElectrodesWidget", I think): ccfwidget, aiohttp
    - the `aiohttp` requirement is a bit of a quirk in that ccfwidget [wasn't installing it's own dependencies properly](https://github.com/NeurodataWithoutBorders/ccf-widget/pull/27) fully, but now that this is fixed upstream, once there is a new PYPI release of ccf, this wouldn't have to be checked here

I think this is ready for testing! Notably, I don't have any real data files with the relevant data fields, so this definitely needs some stress testing to see that it works as expected in those cases. 

Also, depending how bare bones one wants to make the "base" version of the module, there are some additional candidates for making optional, perhaps most optional each of the 3 ndx extensions, which I think _could_ be made optional if the goal is to make the core of `nwbwidgets` lightweight and specifically aligned to match the standard NWB data types. However, these ndx modules seem pretty lightweight, so there is perhaps less to be gained in making them optional (and the potential con of greatly expanding the number of optional dependencies one might have to deal with).

bendichter:
@TomDonoghue wonderful, thanks for putting all of this together. I agree that ndx's could be optional but we might as well keep them as dependencies since they are so lightweight.

You should be able to test out NWB Widgets by streaming files from dandi using ros3. However, plotly isn't working on DANDI Hub right now, which is a separate issue...

setup.py
```diff
@@ -30,6 +30,10 @@
     description="This is nwbwidgets, widgets for viewing the contents of a "
     "NWB-file in Jupyter Notebooks using ipywidgets.",
     install_requires=requirements,
+    extras_require={
+        'human_electrodes' : ['skspatial', 'nilearn', 'trimesh'],
+        'mouse_electrodes' : ['ccfwidget', 'aiohttp']
```
CodyCBakerPhD, line 35:
```suggestion
        'full' : ['ccfwidget', 'aiohttp', 'skspatial', 'nilearn', 'trimesh']
```

We always like to have one of the 'extra' requirements to be an optional full installation if the user might want to be able to view any widget for any neurodata type

CodyCBakerPhD (approved):

PR #225, opened 2022-10-12
add fsspec as stream mode

luiztauffer:
this depends on the next release of pynwb with fsspec
panel_fsspec_stream

### DIFF
--- a/nwbwidgets/panel.py
+++ b/nwbwidgets/panel.py
@@ -9,7 +9,7 @@ import h5py
 
 class Panel(widgets.VBox):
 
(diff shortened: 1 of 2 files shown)


bendichter:
This is great! Let's move the fsspec import inside the if clause though since it isn't installed by default

luiztauffer:
@bendichter should we instead include it in the base requirements.txt? `fsspec` is not heavy and if it takes over ros3 for streaming it might be a good idea to have it as a default installation dependency

bendichter:
@luiztauffer this looks great! I just want to wait until the next pynwb release so that this works out of the box

PR #226, opened 2022-10-17
Docker voila

luiztauffer:
- [x] Container image that serves a Voila process running nwbwidgets with `Panel()`
- [x] Panel update
- [x] documentation update

Depends on #225


bendichter:
lgtm

bendichter:
So how would someone use this?

bendichter:
Oh I see it launches the panel. Could you also allow it to take as input a file path or s3 path in which case it just renders that file?

luiztauffer:
> Oh I see it launches the panel. Could you also allow it to take as input a file path or s3 path in which case it just renders that file?

sure! but then this would be a modification to the Panel, I can make it in a next PR, together with other improvements (e.g. auto-filling the list of existing DANDIsets, etc...)

bendichter:
Would that work? Wouldn't you need to pass a variable to modify the contents of the notebook?

luiztauffer:
The Panel itself should have one extra functionality to be able to read from any S3 path, which is not difficult to implement, it is already doing so indirectly by getting the S3 path from a DANDI set. We wouldn't have to change anything when running the container
Of course, if running from a container this would be mostly used for streaming remote data

bendichter:
would it be possible to make this docker function able to take an s3 path as input? That would allow us to, e.g. create a button on a web interface and launch a viewer for a specific NWB file.

luiztauffer:
@bendichter I changed it now so it works in the following way (see also README):

1. If at container run time we pass an S3 url as an argument, the Notebook will run `nwb2widget` streaming this specific file

2. If no s3 url is passed, it runs with the Panel

bendichter:
using an environmental variable! smart! I'm just going to test this out on my own computer, then it's good to go

bendichter:
When I try to load a local file I'm getting "Invalid local dir path"

luiztauffer:
@bendichter yes I can improve this layout
There are also other necessary improvements on Panel that are easy to make and I'll solve in the coming days:
- Auto-list the existing DANDI sets if DANDI is chosen (only once per run)
- allow for boolean control, on Panel() instantiation, for source options. In a running docker container, supposedly running in a remote machine, the local dir/file options wouldn't make sense, so we could disable these options

if you're running this container, you're getting the local file error because there's no such file inside your container

luiztauffer:
> In a running docker container, supposedly running in a remote machine, the local dir/file options wouldn't make sense, so we could disable these options

that wouldn't be true in the case of running a container which has mounted a virtual file system, which might be the case of DANDI servers! But them it would just be a matter of running Panel with the local_source=True in that case

luiztauffer:
@bendichter 
- added S3 option
- Panel can be instantiated now with boolean controls `enable_dandi_source`, `enable_s3_source` and `enable_local_source`, to control for source data options
- DANDI option auto-lists all DANDI sets in a dropdown, and sows the description of the selected set
- I improved the layout in other small details as well

What I'd also like to do related to this is adding to the documentation:
- Instructions for these new Panel options
- A session for running the container with Voila

but I can also add these in a separate PR later if you'd like to merge this as it is now

Peek 2022-10-24 21-34

docker/README.md
```diff
@@ -0,0 +1,21 @@
...
+```
+
+By default, the Panel will run with `enable_local_source=False` (see [Panel](https://nwb-widgets.readthedocs.io/en/latest/contents/quickstart.html#basic-usage-with-panel)). To allow for local files browsering, run the container with:
+```bash
+$ docker run -p 8866:8866 nwbwidgets-voila -e ENABLE_LOCAL_SOURCE=True
+```
+
+To run for a specific NWB file, pass the s3 url as an ENV variable to the container:
+```bash
+$ docker run -p 8866:8866 nwbwidgets-voila -e S3_URL_NWBFILE=<s3_url_to_nwb_file>
```
bendichter, line 20:
```suggestion
$ docker run -p 8866:8866 -e S3_URL_NWBFILE=<s3_url_to_nwb_file> nwbwidgets-voila 
```

docker/README.md
```diff
@@ -0,0 +1,21 @@
...
+## Run Docker Image
+
+To run with `Panel`:
+```bash
+$ docker run -p 8866:8866 nwbwidgets-voila
+```
+
+By default, the Panel will run with `enable_local_source=False` (see [Panel](https://nwb-widgets.readthedocs.io/en/latest/contents/quickstart.html#basic-usage-with-panel)). To allow for local files browsering, run the container with:
+```bash
+$ docker run -p 8866:8866 nwbwidgets-voila -e ENABLE_LOCAL_SOURCE=True
```
bendichter, line 15:
```suggestion
$ docker run -p 8866:8866 -e ENABLE_LOCAL_SOURCE=True nwbwidgets-voila 
```

PR #232, opened 2022-10-31
[WIP] Trialized time series faceting

h-mayorquin:
The idea is to add a widget for time series that can show trialized data both by filtering (selecting only data where a column -or more- are equal to one value) and faceting (showing the behavior across a grid when filtering for some values).

So far the widget works well as a stand alone:

image

But does not render properly when called from within:
`nwb2widget(nwb)`

@bendichter

### DIFF
--- a/nwbwidgets/timeseries.py
+++ b/nwbwidgets/timeseries.py
@@ -1187,7 +1187,11 @@ class TrializedTimeSeries(widgets.HBox):
         self.trials_table = trials_table
         if self.trials_table is None:
             self.trials_table = time_series.get_ancestor("NWBFile").trials
-                    
+        
+        if self.trials_table is None:
+            self.children = [widgets.HTML("No trials present")]
+            return        
+        
         self.trials_table_df = self.trials_table.to_dataframe()
         
         # Labels to refer to data created by the widget. Should not collapse with the column names on the dynamic table
--- a/nwbwidgets/timeseries.py
+++ b/nwbwidgets/timeseries.py
@@ -1,18 +1,23 @@
 import functools
 from bisect import bisect
 from abc import abstractmethod
+from typing import Union, Optional
 
 import scipy
 import numpy as np
+import pandas as pd
 import matplotlib.pyplot as plt
 
 from ipywidgets import widgets, fixed, Layout
+import plotly.express as px
 import plotly.graph_objects as go
+import plotly.graph_objs as go
 from plotly.subplots import make_subplots
 from plotly.colors import DEFAULT_PLOTLY_COLORS
 
-from pynwb import TimeSeries
 from pynwb.epoch import TimeIntervals
+from pynwb.base import TimeSeries, DynamicTable
+
 
 from .controllers import (
     StartAndDurationController,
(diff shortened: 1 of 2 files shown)


CodyCBakerPhD:
Something I want to add for my data

- [ ] faceting over data columns; it seems fast enough, would be nice way to visualize multiple ROI's at once

Something I need to add for my data

- [ ] TimeInterval table selector; I have multiple tables, and may not always have a trials table


Something to discuss adding would be

- [ ] observers for changes to data column selection; right now I have to push the render button each time, but if I want to rapidly scan through a bunch of ROI's this gets a bit annoying

A widget I'll make that harnesses parts of this is

- [ ] a 'heatmap' view of multiple ROI's plotted over the space of the image/pixel masks

CodyCBakerPhD:
Also just noting that when you use the plotly interact with the faceting it adjusts each plot; is there maybe an option to control that? Or is it just because they all 'share' the x-axis? (basically I would want different x-axis scales each spanning the data range for each condition)

https://user-images.githubusercontent.com/51133164/199066626-4c4dc845-7f96-4b83-aeff-61b3139fa14b.mp4

nwbwidgets/timeseries.py
```diff
@@ -967,3 +972,349 @@ def update(
...
+    else:
+        trials_table_df = trials_table
+
+    if time_series.timestamps:
+        timestamps = time_series.timestamps[:]
+    else:
+        number_of_frames = time_series.data.shape[0]
+        dt = 1.0 / time_series.rate
+        time_offset = time_series.starting_time if time_series.starting_time else 0
+        timestamps = np.arange(0, number_of_frames) * dt + time_offset
```
bendichter, line 992: use https://github.com/NeurodataWithoutBorders/nwbwidgets/blob/230d52e51828493ca141308d42aa91eebc92045d/nwbwidgets/utils/timeseries.py#L8-L41

nwbwidgets/timeseries.py
```diff
@@ -967,3 +972,349 @@ def update(
...
+    data_index_within_trials, index_in_trial_table = np.nonzero(timestamps_contained_in_trial)
+
+    # Check efficiency of accessing HDF5 with indexes
+    data = time_series.data[data_index_within_trials, index]
+    data_df = pd.DataFrame(dict(timestamps=timestamps[data_index_within_trials], data=data))
+
+    # Create an start_time in the data table and merge withe the trial table using a left join operation
+    data_df["start_time"] = [trials_table_df.start_time[index] for index in index_in_trial_table]
+    data_df_trialized = pd.merge(left=data_df, right=trials_table_df, on="start_time", how="left")
+    data_df_trialized["centered_timestamps"] = data_df_trialized.timestamps - data_df_trialized.start_time
```
bendichter, line 1012: can you use https://github.com/NeurodataWithoutBorders/nwbwidgets/blob/master/nwbwidgets/utils/timeseries.py#L244?

bendichter:
@CodyCBakerPhD 

> Also just noting that when you use the plotly interact with the faceting it adjusts each plot; is there maybe an option to control that? Or is it just because they all 'share' the x-axis? (basically I would want different x-axis scales each spanning the data range for each condition)

I don't understand

CodyCBakerPhD:
> I don't understand

If you watch the video, as I attempt to zoom into one of those tiny data plotting regions for a single facet, it also zooms into that region for all the other facets. Likewise, from a zoomed out perspective, looking at all the data for one facet might just leave a tiny blip for the data on another facet. For those cases I'd prefer the plot to span the axis for each facet in a more reasonable way and then indicate the range on the x-ticks

I can also see why you'd want to be able to view everything aligned to the same x-axis for consistency, so maybe this could be a 'lock-axis' option or something

bendichter:
@h-mayorquin it doesn't look like you have a way of controlling the time range. There should be text boxes that allow you to control the time before and after the time of alignment, similar to the PSTH widget

h-mayorquin:
> @h-mayorquin it doesn't look like you have a way of controlling the time range. It looks like in @CodyCBakerPhD 's example you are plotting data for the next 100 seconds? That's way too long. There should be text boxes that allow you to control the time before and after the time of alignment, similar to the PSTH widget

Yes, in the Murthy conversion the trials are very short so is OK. I will see how to add this for the general case though.

Right now the trials are aligned to the beginning of the trial. What do you suggest doing when there is no data before. As in, here there is only data within the trials and they are non-contiguous? My default is returning none in this case.

nwbwidgets/timeseries.py
h-mayorquin: Most of what that function is doing is to prepare a dataframe so the methods of ploly express can be used out of the box (otherwise faceting is way more complicated). The function you point out seems to return a list with the data form the trials, maybe it can indeed substitute a part of it.

h-mayorquin:
I will take a look into the x-axis locking while zooming on faceting thing @CodyCBakerPhD 

@bendichter any idea why would this not render from within the widget accordion but would render on its own?

CodyCBakerPhD:
> any idea why would this not render from within the widget accordion but would render on its own?

I've seen that sometimes if the figure isn't a `go.FigureWidget` or isn't properly otherwise embedded in a `widgets.Output` box, I/we can play around with it tomorrow

nwbwidgets/timeseries.py
h-mayorquin: Done in the last commit.

h-mayorquin:
> @h-mayorquin it doesn't look like you have a way of controlling the time range. It looks like in @CodyCBakerPhD 's example you are plotting data for the next 100 seconds? That's way too long. There should be text boxes that allow you to control the time before and after the time of alignment, similar to the PSTH widget

OK, I did this in the last commit.
@bendichter 
Thinking more about this.
Before this widget ploted the traces within the trial, that is, the traces were plotted from (start_time, to stop_time) . I thought this made sense because start_time and stop_time are the meaningful interval for the trial (after all, that's the reason for them to be in the trials table). 

I thought that the control that made sense was:

`(start_time + "user_selected_time", stop_time + "another_user_selected_time`)`:
  
Where both the user selected times (controlled by widgets) could be positive or negative. 

But looking at the alignment control options in the PSTH:
image

It seems that the plotting interval is:
`("time_chosen_for_alignment" + "user_selected_time", "time_chosen_for_alignment" + "another_user_selected_time`)`:

Where the "time_chosen_for_alignment" is usually the start_time but it can be any other column that follows the convention of ending in `_time`. 

Why does this make more sense than plotting over the whole trial? In what sense was the plotting interval in @CodyCBakerPhD example way too long you mentioned above?

h-mayorquin:
@CodyCBakerPhD 
I addressed your in person comment of disabling the filtering selection whe not usted in the last commit.
I think I can disable the x-locking behavior but I don't think it is a good idea to disable it in general. Plotly documentation advice against it and I guess they made it a default for a reason:
https://plotly.com/python/facet-plots/

We can have a checkbox that controls for whether the behavior is active or not if you and @bendichter  think that is desirable.

h-mayorquin:
State of the widget as the last commit:

https://user-images.githubusercontent.com/6429509/199577893-b167cb45-449f-4eb2-83fe-03dbdb274c36.mp4

CodyCBakerPhD:
> Plotly documentation advice against it and I guess they made it a default for a reason:
https://plotly.com/python/facet-plots/

> We can have a checkbox that controls for whether the behavior is active or not if you and @bendichter think that is desirable.

Yeah an optional toggle is all I'm asking for there. Common axes are a good default, but in cases where some slices are very thing on certain facets it can make more sense to try to expand different axes to note the trends one might see on the zoomed-in scale

nwbwidgets/timeseries.py
h-mayorquin: I ended up adding a vectorized version of the function that bisects in the utils.

h-mayorquin:
The widget so far:

https://user-images.githubusercontent.com/6429509/199812731-992ed3a2-b0e1-467e-a4d2-ed24cbc693cc.mp4

CodyCBakerPhD:
@h-mayorquin If you want more examples to test/develop with, here's a snippet for using my latest files (recommend using on DANDI Hub for optical bandwidth speeds)

```python
import h5py
import fsspec
from fsspec.implementations.cached import CachingFileSystem
from pynwb import NWBHDF5IO
from nwbwidgets.timeseries import TrializedTimeSeries

s3_url = "https://dandiarchive.s3.amazonaws.com/blobs/297/fd6/297fd6b9-4c28-48c4-8c86-e9ef2a0fb07d"
cfs = CachingFileSystem(
    fs=fsspec.filesystem("http"),
    cache_storage="/home/jovyan/fsspec_cache",  # Local folder for the cache
)
file_system = cfs.open(s3_url, "rb")
file = h5py.File(file_system)
io = NWBHDF5IO(file=file, load_namespaces=True)
nwbfile = io.read()

# Choose series
#roi_response_series = nwbfile.processing["ophys"]["DfOverF"].roi_response_series["NeuronDfOverF"]
roi_response_series = nwbfile.processing["ophys"]["Fluorescence"].roi_response_series["NeuronFluorescence"]

# Choose table
#target_table = nwbfile.processing["behavior"]["SwimIntervals"]
target_table = nwbfile.processing["behavior"]["ActivityStates"]  # Causes issue due to instantaneous states

TrializedTimeSeries(time_series=roi_response_series, trials_table=target_table)
```

h-mayorquin:
OK, I had to do some changes but now it renders with `nwb2widget(nwb)`:

 

https://user-images.githubusercontent.com/6429509/200915824-4c7eeab1-a6ea-4778-962d-99416560d248.mp4

bendichter:
@h-mayorquin great! What did you need to change?

h-mayorquin:
> @h-mayorquin great! What did you need to change?

There was a silent error because the `nwb2widget` function requires neurodata_vis_spec items to have a `neurodata_vis_spec` as an argument if they are callable. I added a router function (`route_trialized_time_series`) with this argument.

h-mayorquin:
@CodyCBakerPhD I added a toggle to control the x-locking behavior in the last commit. Could you test it to confirm that is the behavior that you wanted?

nwbwidgets/ophys.py
```diff
@@ -105,13 +105,7 @@ def update_figure(index=0):
...
 
 def show_df_over_f(df_over_f: DfOverF, neurodata_vis_spec: dict):
-    if len(df_over_f.roi_response_series) == 1:
-        title, data_input = list(df_over_f.roi_response_series.items())[0]
-        return neurodata_vis_spec[RoiResponseSeries](
-            data_input, neurodata_vis_spec, title=title
-        )
-    else:
-        return neurodata_vis_spec[NWBDataInterface](df_over_f, neurodata_vis_spec)
+    return neurodata_vis_spec[NWBDataInterface](df_over_f, neurodata_vis_spec)
```
h-mayorquin, line 108: There is a problem here when we make the `neurodata_vis_spec[RoiResponseSeries` an OrderedDict to include the trialized widget as an option (see the behavior in the `view.py` module). 

Because `show_df_over_f` is a callable the function `nwb2widget` expects this output to be a visualization (not an ordered dict). The simplest solution was to roll back the `show_df_over_f` to return the visualization for the whole `Container/DataInterface` and then the visualization  can be properly display as a root of a `lazy_tab`. That is how it is working right now.

The other option was to modify  `nwb2widget` to handle the case when callables return `OrderDict` but that seemed like a larger modification given that `nwb2widget` is such a central function and this might have non-tested side effects in other visualizations. I opted for the simple solution here as that only touches the visualizations concerning the RoiExtractors and DfOverF objects.

h-mayorquin:
Testing widget upload
https://user-images.githubusercontent.com/6429509/200926070-397302df-01dc-4180-afe6-38be3134a39e.mp4

h-mayorquin:
Again:

https://user-images.githubusercontent.com/6429509/200926440-8ab7eed2-8b7a-4a6d-9805-0f292c356485.mp4

nwbwidgets/timeseries.py
```diff
@@ -966,4 +942,485 @@ def update(
         fig = fig if fig is not None else go.FigureWidget()
         fig.data = []
         fig.layout = {}
-        return self.plot_group(group_inds,data,time_ts_aligned,fig,order)
+        return self.plot_group(group_inds, data, time_ts_aligned, fig, order)
+
+
+def get_time_series_data(
```
bendichter, line 948: Could you use this function and adjust it as necessary?

https://github.com/NeurodataWithoutBorders/nwbwidgets/blob/dba14008af7f7aea6f3ea83f010a0a89cca1bfa4/nwbwidgets/utils/timeseries.py#L86

nwbwidgets/timeseries.py
```diff
@@ -966,4 +942,485 @@ def update(
...
+        conversion_factor_scalar = time_series.conversion
+        channel_conversion_vector = (
+            time_series.channel_conversion if hasattr(time_series, "channel_conversion") else np.ones_like(data)
+        )
+
+        data = data * channel_conversion_vector * conversion_factor_scalar + offset_scalar
+    return data.flatten()
+
+
+def trialize_time_series(
```
bendichter, line 973: Could you use this function?

https://github.com/NeurodataWithoutBorders/nwbwidgets/blob/dba14008af7f7aea6f3ea83f010a0a89cca1bfa4/nwbwidgets/utils/timeseries.py#L266

nwbwidgets/timeseries.py
```diff
@@ -966,4 +942,485 @@ def update(
...
+                "showarrow": False,
+                "font": {"size": 28},
+            }
+        ],
+    )
+
+    return empty_figure
+
+
+def calculate_moving_average_over_trials(df, moving_average_window):
```
bendichter, line 1044: This seems specific to your widget. Maybe it would be better as a static method.

nwbwidgets/timeseries.py
```diff
@@ -966,4 +942,485 @@ def update(
...
+
+    df_sort = df.sort_values(by="centered_timestamps")
+    df_sort["moving_average"] = df_sort["data"].rolling(moving_average_window).mean()
+
+    return df_sort
+
+
+def add_moving_average_traces(figure, df, facet_col, facet_row):
```
bendichter, line 1052: This seems specific to your widget. Maybe it would be better as a static method.

nwbwidgets/timeseries.py
```diff
@@ -966,4 +942,485 @@ def update(
...
+                    showlegend=False,
+                    line=dict(color="black", width=4),
+                    row=row_index + 1,
+                    col=col_index + 1,
+                )
+
+    return figure
+
+
+def build_faceting_figure(df, facet_col, facet_row, data_label="data", trial_label="trial"):
```
bendichter, line 1153: This seems specific to your widget. Maybe it would be better as a static method.

nwbwidgets/timeseries.py
```diff
@@ -966,4 +942,485 @@ def update(
...
+        figure = build_faceting_figure(df=self.trialized_data_df, facet_col=facet_col, facet_row=facet_row)
+
+        # Update the widget
+        with self.figure_widget.batch_update():
+            self.figure_widget.update(layout_annotations=None)
+            self.figure_widget.update(figure.to_dict(), overwrite=True)
+            self.figure_widget.update_xaxes(matches=matches)
+
+
+def route_trialized_time_series(time_series: TimeSeries, neurodata_vis_spec=None, **kwargs):
```
bendichter, line 1416: There is no routing going on here..

bendichter: It would also be good if you subclassed `TrializedTimeSeries` to have proper labels for specific modalities

h-mayorquin: I was thinking that would be the role of the routing function. Initializing the visualization -for lack of a better term- to match the neurodata type that is specified for it in the visualization spec dictionary.

nwbwidgets/timeseries.py
h-mayorquin: Most of the function implemented there is to build the dataframe into the form where the faceting functionality of plotly can be used.

There is a place inside `trialized_time_series` where the data from trials is extracted and also the timestamps. For that part, after your last suggestion above, I took a look a the functions in that utils module for timeseries. I don't think they fit that well here without introducing more unnecessary computation. 

Basically, right now I have a vectorized bisect function (as provided by numpy) which I use to extract the set of trialized indexes and then I use them on both the data and the timestamps. The functions in the utils module are non vectorized and they return the data in a list without the indexes which I need for the timestamps. Using `align_by_time_intervals`, for example, would reduce the performance here both by losing the vectorization and then by requiring some more computation to get the indexes of the timestamps.

I think the generalization would be to have a general vectorized function that return the trialized indexes. That hypothetical function then can be used both here and inside `bisect_timeseries_by_times` to avoid the per trial bisection there. 

What do you think?

nwbwidgets/timeseries.py
h-mayorquin: I wanted to separate the logic of plot creation (here in the function `build_faceting_figure`) from the logic of the widget. In my view this separation of concerns (composition) makes it easier to modify and test the plotting function. Reason being:
1) It is easier for someone who does not know the function to get the plotting function and just use it by itself to understand its behavior. In opposition to this, in the `AlignMultiTraceTimeSeriesByTrialsVariable` you need to go very deep in the function to understand what is the plotting function.
2) The plotting function can be tested or modified with minimal modification to the widget class. I think keeping the points of contact between the plotting function and all the other widget class responsibilities small leads to better design. 

Basically I view the widget class a place with the following responsabilities:
1) Data loading from the appropriate neuro data type.
2) Control definition (maybe can be done with composition using your concept of foreign controlers).
3) Handling the interaction between controls and the (external) plotting function.

What do you think? Are there downsides that I am not seeing for this separation?

nwbwidgets/timeseries.py
h-mayorquin: On it.

h-mayorquin: Done.

nwbwidgets/timeseries.py
h-mayorquin: That said, I can create a plotting class and make all the methods that are floating now by themselves methods of that class to give the code more coherence.

bendichter: I think a better way to organize would be to separate calculation from plotting.

h-mayorquin: Thanks, this got me thinking:
#239

h-mayorquin:
> @h-mayorquin the sorting looks right. Good thinking using natsort. Could you show a screenshot that displays this working correctly?

Here is some recording of it working on the test data that I have been using:

https://user-images.githubusercontent.com/6429509/204795830-4e558c62-df9c-4a89-b843-fc22fa39e40f.mp4

h-mayorquin:
@bendichter 
I ordered the categories so they appear ordered when facetting.
image

nwbwidgets/timeseries.py
h-mayorquin: This is done in the last commit and in the routing function.

bendichter (approved):

PR #233, opened 2022-11-01
modularize docs

bendichter:
* separate docs into modules
* formal links to autogen API of this library and of pynwb


luiztauffer (approved):
looks good!

Issue #239, opened 2022-11-30
Proposal for organizing VisualizationWidgets


h-mayorquin:
This is a draft / work in progress. As we are finishing #232 and I am switching to work in other projects I wanted to write my thoughts of what could be a general schema or abstract class for the visualization widgets in the library. Two preliminary points:

1) I am not sure yet if going all the way to an abstract class at this point is useful (or if it will be ever be). Visualization is a category that is difficult to formalize and maybe over-constraining the design with an abstract class will create more problems that it solves. So the proposal here starts more humble as a mere suggestion for code organization. 
2) This is specific to the widgets that contain some sort of plot. Not all of them do, as an example,  [nwbwidgets.file.show_nwbfile](https://nwb-widgets.readthedocs.io/en/latest/contents/neurodata_widgets.html#session-metadata) just displays some text and is outside of the scope of the proposal here.

## Proposal
My claim is that is useful to think about visualization widgets as coordination classes in charge composition, layout and interaction between controllers, data and plotting. The idea can be sumarized in the following diagram:
![2022-11-30(VisualizationWdiget)](https://user-images.githubusercontent.com/6429509/204785284-bc5eaf47-3c81-444e-96c0-02843f1e7de8.png)

In brief, the role of the `VisualizationWidget` is to coordinate the interaction between a `ControllerClass`, a `CanvasClass` and the data extracted and processed from the neurodata types (i.e. a TimeSeries or a DynamicTable). A minor role of the `VisualizationWidget` is to define the layout between the controllers and the canvas (usually done with [Box](https://ipywidgets.readthedocs.io/en/7.6.3/examples/Widget%20Styling.html#Natural-sizes,-and-arrangements-using-HBox-and-VBox) mechanism). The most important responsability of the `VisualizationWidget` is to define the interactions between the data and the canvass which is done with the [ jupyter widgets observer mechanism](https://ipywidgets.readthedocs.io/en/7.6.2/examples/Widget%20Events.html#Traitlet-events).  Furthermore, I think we should aim to isolate the definition of the interactions from the computational and rendering actions with two general methods as shown in the following diagram:
![2022-11-30(VisualizationWidgetDetails)](https://user-images.githubusercontent.com/6429509/204788943-4e0141ef-9c1f-4565-aa55-9eafa3ed9ac0.png)

[...]

PR #241, opened 2022-12-04
Refactor TwoPhotonSeries

CodyCBakerPhD:
Inspired to test the proposal structure of #239, and replacing #230, I refactored the `ophys` submodule to follow the proposed workflow to show how much more readable the code is, as well as how well it can generalize to allow easy creation of subwidgets that differ only slightly in how they setup different parts of the total visualization.

The TwoPhotonSeries visualizations are a good case to use for this because they use essentially the same types of plots for planar view of both single-plane instances (3D data) and volumetric cases (4D data). Thus if any improvements are ever made to either plot, if they didn't have some type of inheritance structure those improvements would have to be duplicated in code. Also we have several project going on right now and coming up (looking at MICrONS) that use these modalities, so further enhancements are no doubt to come.

Several such improvements have been made here...
- the addition of a rotation feature allows the user to choose a different axis of orientation
- contrast levels can be sensitive to different source data; imshow allows either automated methods (these are just 'OK' I'd say) or manual specification of ranges
- the contrast options were starting to add clutter to the window, so for users that aren't looking to control that (and of course, by default) these are all hidden under a 'Simplified' view toggle, which can then be expanded. This is especially useful if we decide to eventually add a faceting, or possibly downsampling, filtering, or coloration controllers to allow maximum user control over the configuration of the call to `plotly.express.imshow`

Both the `SinglePlaneVisualization` and the `PlaneSliceVisualization` share these new features - the only difference between them is the latter has a single extra controller to allow specification of the plane index. As such, the inherited class does as little as possible to produce the key differences.

I also tried to keep all controllers here as general as possible; only the final assembly-level ones are kept in the `ophys_controllers.py` as being specifically designed for use the other `ophys` classes. This includes the controllers which have multiple components and their own internally interacting observers, which is one of the design proposals from @h-mayorquin that would up helping a lot to reduce complexity.

Another really helpful suggestion from @h-mayorquin was the `MultiController` class for assembling these components and allowing an outer exposure of the attributes so we don't have to meticulously (and painfully, sometimes) recurse down through the levels of `ipywidgets.Box.children` to find an object we want to reference. It also makes these references more verbose, and thus the code more readable.

I also re-organized the file structure to be more readable/navigable instead of everything being in a single file. When you start piling on more visualization for neurodata types from a single modality like that, it gets hard to scroll through looking for items you might want to find; even ctrl-F or GitHub link navigation doesn't always help much.

If this is too much to review all at once, I'd be happy to break the individual controllers into modular PRs accompanied by their own tests and even documentation. Figured it might be nice to see an explicit example of proposal structure #239 in practice and how it can help us keep this project more easily manageable going forward.

Video demonstration:

https://user-images.githubusercontent.com/51133164/205493069-e894bb0e-07d9-4003-9172-c7dd877c8ac1.mp4

TODO:
- [ ] need to add tests for various modular components


nwbwidgets/ophys/segmentation.py
```diff
@@ -356,16 +236,6 @@ def route_plane_segmentation(plane_seg: PlaneSegmentation, neurodata_vis_spec: d
         return PlaneSegmentation2DWidget(plane_seg)
 
 
-def show_grayscale_volume(vol: GrayscaleVolume, neurodata_vis_spec: dict):
-    import ipyvolume.pylab as p3
-
-    fig = p3.figure()
-    p3.volshow(vol.data, tf=linear_transfer_function([0, 0, 0], max_opacity=0.1))
-    return fig
```
CodyCBakerPhD, line 364: This function wasn't doing anything different from the usual volume shower above, so remapped the extension neurodatatype in the `viz_spec` to point to the other one instead

nwbwidgets/ophys/single_plane.py
```diff
@@ -0,0 +1,202 @@
...
+        HDF5 chunks?
+        """
+        itemsize = self.two_photon_series.data.dtype.itemsize
+        nbytes_per_image = math.prod(self.two_photon_series.data.shape) * itemsize
+        if nbytes_per_image <= max_mb_treshold:
+            self.update_data(frame_index=0)
+        else:
+            # TOD: Figure out formula for calculating by in one-shot
+            by_width = 2
+            by_height = 2
```
CodyCBakerPhD, line 89: Still a TODO item

nwbwidgets/ophys/two_photon_series.py
```diff
@@ -0,0 +1,62 @@
...
+                self.controls = dict(slider=self.slider)
+                self.slider.observe(lambda change: update_figure(index=change.new), names="value")
+
+                update_figure()
+
+                series_name = indexed_timeseries.name
+                base_title = f"TwoPhotonSeries: {series_name}"
+                self.figure.layout.title = f"{base_title} - read from first external file"
+
+                self.children = [self.figure, self.slider]
```
CodyCBakerPhD, line 50: TBH this entire section could also be absorbed into a minimal child class (the only thing different is where/how the data is fetched...

Might prefer to do in separate PR though, this is more than enough as is.

nwbwidgets/ophys/single_plane.py
```diff
@@ -0,0 +1,202 @@
...
+            self.data = self.two_photon_series.data[0, ::by_width, ::by_height]
+
+    def get_rotation(self) -> int:
+        """The rotation attribute of the SinglePlaneDataController cannot be attached in a modifiable state."""
+        if not hasattr(self, "Controller"):  # First time this is called
+            return 0
+        return self.Controller.components[self.data_controller_name].components["RotationController"].rotation
```
CodyCBakerPhD, line 96: This is the only time a reference to a nested child is necessary (due to attribute mutability relative to the button observer events)

Technically I could override those observers and remap it all to an attribute specific to these classes, but that ends up taking many more lines of code than this simple router

That said, I built the `MutliController` to at least have dictionary mapping capability, so you can navigate it entirely based on name references to the components (much easier than children indexes)

nwbwidgets/controllers/multicontroller.py
```diff
@@ -0,0 +1,32 @@
...
+            if isinstance(component, widgets.Widget) and not isinstance(component, MultiController):
+                children.append(component)
+
+        self.children = tuple(children)
+        self.controller_fields = tuple(controller_fields)
+
+        self.setup_observers()
+
+    def setup_observers(self):
+        pass
```
bendichter, line 32: I don't like the distinction between Controllers and MultiControllers. I think a widget that contains multiple controllers should just be another controller.

nwbwidgets/ophys/ophys_controllers.py
```diff
@@ -0,0 +1,97 @@
+from typing import Optional
+
+import ipywidgets as widgets
+
+from ..controllers import RotationController, ImShowController, ViewTypeController, MultiController
+
+
+class FrameController(widgets.VBox):
```
bendichter, line 8: There should be a `Controller` base class.

nwbwidgets/ophys/ophys_controllers.py
```diff
@@ -0,0 +1,97 @@
...
+            max=1,  # Actual value will depend on data selection
+            orientation="horizontal",
+            description="Frame: ",
+            continuous_update=False,
+        )
+
+        self.children = (self.frame_slider,)
+
+
+class PlaneController(widgets.VBox):
```
bendichter, line 26: I think defining controllers that contain a single widget is a bit overkill. In my mind, the point of controllers is collections of widgets in a system that is reusable.

nwbwidgets/ophys/ophys_controllers.py
```diff
@@ -0,0 +1,97 @@
...
+
+class SinglePlaneDataController(MultiController):
+    def __init__(self, components: Optional[list] = None):
+        default_components = [RotationController(), FrameController()]
+        if components is not None:
+            default_components.extend(components)
+        super().__init__(components=default_components)
+
+        # Align rotation buttons to center of sliders
+        self.layout.align_items = "center"
```
bendichter, line 52: why are you allowing this to be extended with more components?

nwbwidgets/ophys/ophys_controllers.py
```diff
@@ -0,0 +1,97 @@
+from typing import Optional
+
+import ipywidgets as widgets
+
+from ..controllers import RotationController, ImShowController, ViewTypeController, MultiController
+
+
+class FrameController(widgets.VBox):
+    controller_fields = ("frame_slider",)
```
bendichter, line 9: maybe this should be a dictionary, which maps a class attribute to an internal widget or controller value. e.g.

```python
controller_fields = dict(frame: "frame_slider")
```

would imply that frame_controller.frame gives you the `value` of `frame_controller.frame_slider`.

nwbwidgets/ophys/segmentation.py
```diff
@@ -2,117 +2,33 @@
...
+from ..timeseries import BaseGroupedTraceWidget
+from ..utils.cmaps import linear_transfer_function
+from ..utils.dynamictable import infer_categorical_columns
+from ..controllers import ProgressBar
 
 
 color_wheel = px.colors.qualitative.Dark24
 
 
-class TwoPhotonSeriesWidget(widgets.VBox):
```
bendichter, line 34: You took this entire class out? Can you handle all the use cases, e.g. reading from an external file?

nwbwidgets/controllers/multicontroller.py
```diff
@@ -0,0 +1,32 @@
+from typing import Tuple, Dict
+
+import ipywidgets as widgets
+
+
+class MultiController(widgets.VBox):
```
bendichter, line 6: Can you make this inherit from `widgets.Box`? See 

https://ipywidgets.readthedocs.io/en/7.x/examples/Widget%20Styling.html#The-VBox-and-HBox-helpers

CodyCBakerPhD: Yes I started running into this on another PR as well

nwbwidgets/controllers/multicontroller.py
CodyCBakerPhD: Yeah, I just need to utilize a base Controller class, which a MultiController itself is as well in every respect

nwbwidgets/ophys/segmentation.py
CodyCBakerPhD: It's actually just renamed: https://github.com/NeurodataWithoutBorders/nwbwidgets/pull/241/files#diff-5bf5b8acc0948de1db1c627366b4eed9ac2b1965bbaf82ca66dc3170992c3444R13

The exact same structure as before is in place, it just deploys to subclass visualizations in a more organized way to allow better scalability (easier to go in and add a quick feature to a single subclass than pile on more lines in a common class that way)

CodyCBakerPhD:
Going to modularize certain aspects of the PR, like the controllers, into separate sub-PRs. Moving to draft for now

PR #244, opened 2022-12-14
Add pre commit configs

CodyCBakerPhD:
@bendichter Initial setup for enabling pre-commit. Once this is on main I can turn the bot on and then go through the other PRs to fix any conflicts


CodyCBakerPhD:
It appears that several tests have been failing since the merge of #232 

image

This also seems to have been overlooked because they never even triggered on that PR - which actually seems true of all external forks. How should we handle this?

bendichter (approved):

PR #246, opened 2022-12-16
Fix CI not running on external forks

CodyCBakerPhD:
CI hasn't been triggering on forks, which led to an issue in #232 causing a bunch of tests to break. Fixes for those are upcoming from @h-mayorquin 

The core of the issue is the trigger condition was `push`, which only happens relative to the host repo during the final merge commit. You always want to use `pull_request` to allow trigger from outside parties; +/- it should give a prompt for first-time contributors for safety reasons, and they won't ever have access to private repo action secrets

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -1,5 +1,8 @@
 name: Tests and coverage
-on: [push]
+
+on:
+  pull_request:
+
 jobs:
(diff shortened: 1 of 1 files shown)


CodyCBakerPhD:
Of course, can't test this until it's on `master` branch and I try an external fork from that...

PR #247, opened 2022-12-16
Fix circular imports

CodyCBakerPhD:
`isort` turns out to be pretty useful - not only do we never have to worry about following best practices for import ordering again, it also revealed a number of circular imports that were technically bypassed by way of absolute imports from the package namespace


CodyCBakerPhD:
@bendichter 

@h-mayorquin has identified the source problems and is working on a fix to the rest of the tests that branches from this PR

bendichter (approved):

PR #248, opened 2022-12-16
Enhance issue forms

CodyCBakerPhD:
Enhanced issue report forms, particularly a bug report that automatically asks people for the package version they are using


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,89 @@
...
+    id: reproduce
+    attributes:
+      label: Steps to Reproduce
+      description: |
+        Please copy and paste the code you were trying to run that caused the error.
+
+        Feel free to include as little or as much as you think is relevant. This section will be automatically formatted into code, so no need for backticks.
+      render: python
+    validations:
+      required: true
```
bendichter, line 29:
```suggestion
```

bugs don't always have tracebacks.  Sometimes the code just does not act as expected but does run, so I don't think this field should be required.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,89 @@
...
+  - type: textarea
+    id: reproduce
+    attributes:
+      label: Steps to Reproduce
+      description: |
+        Please copy and paste the code you were trying to run that caused the error.
+
+        Feel free to include as little or as much as you think is relevant. This section will be automatically formatted into code, so no need for backticks.
+      render: python
```
bendichter, line 27:
```suggestion
```

I don't like assuming this. I often want to add plain text here, or show code in other languages e.g. bash

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,89 @@
...
+      required: true
+  - type: dropdown
+    id: executable
+    attributes:
+      label: Python Executable
+      options:
+        - Conda
+        - Python
+    validations:
+      required: true
```
bendichter, line 55: Is this helpful information? I don't think I have ever used this in a debug.

CodyCBakerPhD: For widgets they definitely have less options, we can skip it

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,87 @@
...
```
CodyCBakerPhD, line 55:
```suggestion
```

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,78 @@
...
+    validations:
+  - type: textarea
+    id: traceback
+    attributes:
+      label: Traceback
+      description: |
+        Please copy and paste the traceback produced by the error.
+
+        This section will be automatically formatted into code, so no need for backticks.
+      render: python
```
bendichter, line 36:
```suggestion
      render: pytb
```

A Python traceback should be rendered as pytb

CodyCBakerPhD:
@bendichter Everything is optional now except Platform/Python Versions and signing the Code of Conduct / non-duplicate issue

bendichter (approved):

PR #249, opened 2022-12-16
Add auto-publish workflow

CodyCBakerPhD:
Surprised this wasn't already on here - added a workflow to automatically publish to PyPI on package release


bendichter (approved):

CodyCBakerPhD:
Token is added! Should work, but we'll keep an eye on the action next time we do a release just to make sure

PR #250, opened 2022-12-16
Fix remaining tests

h-mayorquin:
This is a PR following some tests that were brocken with #232 and some small mistakes corrected after enabling isort.

For the former, they have to do with the shape of the what the  `get_timeseries_in_units` is returning (I was flatttening the input not realizing that sometimes you want the whole `TimeSeries` and not just a specific column) I corrected this so the old tests pass.

There was a problem with a circular import as the function name `nwb2widget` is duplicated in view and base. I used a named import to solve this, let me know if you think about a better solution @CodyCBakerPhD.

Finally, I needed to ammend a small test as now the `df_over_f` returns a different widget (the one in #232).

### DIFF
--- a/nwbwidgets/view.py
+++ b/nwbwidgets/view.py
@@ -9,7 +9,6 @@ from ipywidgets import widgets
 from ndx_icephys_meta.icephys import SweepSequences
 from ndx_spectrum import Spectrum
 
-from .base import nwb2widget as nwb2widget_base
 from .base import (
     dict2accordion,
     processing_module,
@@ -19,6 +18,7 @@ from .base import (
     show_multi_container_interface,
     show_neurodata_base,
     show_text_fields,
+    nwb2widget as nwb2widget_base,
 )
 
 from .behavior import route_spatial_series, show_behavioral_events
--- a/nwbwidgets/view.py
+++ b/nwbwidgets/view.py
@@ -9,84 +9,109 @@ from ipywidgets import widgets
 from ndx_icephys_meta.icephys import SweepSequences
 from ndx_spectrum import Spectrum
 
-from . import (
-    base,
-    behavior,
-    ecephys,
-    file,
-    icephys,
-    image,
-    misc,
-    ophys,
-    spectrum,
-    timeseries,
+from .base import (
+    dict2accordion,
+    processing_module,
+    render_dataframe,
+    show_dset,
+    show_fields,
+    show_neurodata_base,
+    show_text_fields,
 )
-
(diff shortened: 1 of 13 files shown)


nwbwidgets/view.py
```diff
@@ -118,4 +120,4 @@ def show_dynamic_table(node, **kwargs) -> widgets.Widget:
 
 
 def nwb2widget(node, neurodata_vis_spec=default_neurodata_vis_spec):
-    return base.nwb2widget(node, neurodata_vis_spec)
+    return nwb2widget_base(node, neurodata_vis_spec)
```
CodyCBakerPhD, line 123: Ah, good catch. Unfortunate naming convention

h-mayorquin:
The error appearing here does not appear on my system. I think is a dependency issue. I will test it in a clean env.

test/test_ecephys.py
```diff
@@ -102,5 +102,5 @@ def test_show_spectrogram():
     show_spectrogram(ts, channel=channel)
 
 
-def test_show_ccf():
-    show_ccf()
+# def test_show_ccf():
+#     show_ccf()
```
h-mayorquin, line 106: @bendichter @CodyCBakerPhD 
I fixed all the other tests related to the changes that I made. Initially I thought that this remaining test was broken due to the changes in #219.

However, I just tried running:
```python
from ccfwidget import CCFWidget
ccf = CCFWidget()
ccf
```
(As indicated in https://github.com/NeurodataWithoutBorders/ccf-widget)
In a clean install with fre-dependencies here and it I get an the same error that was appearing here. So it seems that the problem is with `ccfwidget`.  

I have a hard time seeing how this relates to any of the changes that I did as the test just runs with empty arguments mimicing the run instructions in the repo.

h-mayorquin (commented):
I think that everything save the test in the comment below is fixed. In my opinion, we should merge this have a working build and open and issue for this.

What do you think?

test/test_ecephys.py
CodyCBakerPhD: Yeah I'm looking into it now. It's odd because no actual behavior seems to have changed since the last time the tests ran successfully

CodyCBakerPhD:
Thanks for figuring it out @h-mayorquin, I'll take it from here and if I can't get to the bottom of it then yeah we'll just to a `pytest.skip` on that test with a note that it broke at some point for some reason

CodyCBakerPhD (approved):

PR #256, opened 2023-01-16
Add BaseController - demonstrate refactor on time window controllers

CodyCBakerPhD:
Draft PR split from #255, consists of 4 major parts:

(i) Introduction of `BaseController`

(ii) Introduction of `MultiController`

(iii) Demonstrated refactor of the 3 classes in the time window controllers to inherit from the `BaseController`, which thereby allows them to be passed as a component to a `MultiController`. Also added annotation types.

(iv) Tests for new controller classes - still in progress until I get the go ahead to finish them up, but a few of them show the new `BaseController` design pattern.

I opted to get rid of the `GenericController` from #255 since it was allowing a bit too much flexibility to controller design; better if we just enforce a single design standard that is similar to existing controller designs.


nwbwidgets/controllers/basecontroller.py
```diff
@@ -0,0 +1,81 @@
...
+
+        To align the children vertically, set `your_controller.layout.flex_flow = "column"`.
+        """
+        super().__init__()  # Setup Box properties
+        self.layout.display = "flex"
+        self.layout.align_items = "stretch"
+
+        self.setup_attributes()
+        self.setup_children()
+        self.setup_observers()
```
CodyCBakerPhD, line 38: This 3-step process seems to describe most controllers quite well, with enough flexibility in inheriting/overriding any (including `__init__` as long as `super()` is called) to keep current behavior the same

nwbwidgets/controllers/basecontroller.py
```diff
@@ -0,0 +1,81 @@
...
+    a particular child.
+    """
+
+    def __init__(self):
+        """
+        Defines the workflow for initializing a Controller.
+
+        Overriding this method in a child class without calling super().__init__() is highly discouraged.
+
+        To align the children vertically, set `your_controller.layout.flex_flow = "column"`.
```
CodyCBakerPhD, line 30: I did notice  in the time window controllers, many take an input `orientation: Literal["horizontal", "vertical"]` - is this something you'd like added to the base class, potentially?

Right now I'm leaving it entirely in the hands of the user to modify the `.layout` after initializing the controller

nwbwidgets/controllers/time_window_controllers.py
```diff
@@ -1,101 +1,102 @@
...
-    # @property
-    # def value(self):
-    #    """Must be window in seconds"""
-    #    pass
-
-    @abstractmethod
-    def get_children(self):
-        pass
-
-    @abstractmethod
```
CodyCBakerPhD, line 48: Note that the usage of `abstractmethod` here does nothing since the class does not inherit from `ABC`. Normally this would prevent instantiation of the class entirely with an informative error telling the user to define those methods.

Swapped to raising `NotImplementedError`s on function call instead.

nwbwidgets/controllers/time_window_controllers.py
```diff
@@ -1,101 +1,102 @@
...
 
-class WindowController(HBox, ValueWidget, DescriptionWidget):
+
+class WindowController(BaseController, ValueWidget, DescriptionWidget):
     def __init__(
-        self, vmin, vmax, start_value=(None, None), description="window (s)", orientation="horizontal", **kwargs
+        self,
+        vmin: int,
+        vmax: int,
+        start_value: Optional[Tuple[Optional[int], Optional[int]]] = (None, None),
```
CodyCBakerPhD, line 14: Not so sure about this type annotation; this class is the only one that initializes it with `(None, None)`, the others take `None` and figure out tuple contents from other values

nwbwidgets/controllers/time_window_controllers.py
```diff
@@ -1,101 +1,102 @@
...
-        pass
+        raise NotImplementedError("The 'move_end' method of this WindowController has not been defined.")
+
+    def setup_observers(self):
+        self.to_start_button.on_click(self.move_start)
+        self.backwards_button.on_click(self.move_down)
+        self.forward_button.on_click(self.move_up)
+        self.to_end_button.on_click(self.move_end)
+
+    def get_children(self) -> widgets.Box:
```
CodyCBakerPhD, line 61: I'm not entirely sure if this method is ever used in a static way, though as written it certainly could.

I had originally thought of controllers and their methods as modifying things entirely in-place

test/test_controllers/test_multicontroller.py
```diff
@@ -0,0 +1,147 @@
...
+        self.children = (self.button, self.check_box)
+
+    def check_box_on_button_click(self, change):
+        if self.check_box.value is True:
+            self.check_box.value = False
+        elif self.check_box.value is False:
+            self.check_box.value = True
+
+    def setup_observers(self):
+        self.button.on_click(self.check_box_on_button_click)
```
CodyCBakerPhD, line 28: This block shows the intended design of most basic controllers using the 3-step pattern.

For the most part, overriding `__init__` is not necessary at all, just

(a) define your attributes

(b) define how you want the children to be arranged (could also consist of nested boxes if needed)

(c) define observers on the attributes (which could be methods of the class or locally defined ones in `setup_observers`)

PR #260, opened 2023-01-31
Relax upper bound on ipywidgets

CodyCBakerPhD:
@bendichter Satra needs looser bounds on ipywidgets to get the Hub rebuild of default kernels to work smoothly

The versions above `ipywidgets>=8.0.0` fix the javascript issue

### DIFF
--- a/nwbwidgets/base.py
+++ b/nwbwidgets/base.py
@@ -182,7 +182,7 @@ def lazy_show_over_data(list_, func_, labels=None, style: GroupingWidget = widge
     children = [vis2widget(func_(list_[0]))] + [widgets.HTML("Rendering...") for _ in range(len(list_) - 1)]
     out = style(children=children)
     if labels is not None:
-        [out.set_title(i, label) for i, label in enumerate(labels)]
+        [out.set_title(i, str(label)) for i, label in enumerate(labels)]
 
     def on_selected_index(change):
         if change.new is not None and isinstance(change.owner.children[change.new], widgets.HTML):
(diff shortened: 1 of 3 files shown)


bendichter (dismissed):

CodyCBakerPhD:
Trying to debug the trailets issue in the tests now, didn't prohibit me from using the panel or `nwb2widget` in practice

bendichter:
Huh, good to know

nwbwidgets/base.py
```diff
@@ -182,7 +182,7 @@ def lazy_show_over_data(list_, func_, labels=None, style: GroupingWidget = widge
     children = [vis2widget(func_(list_[0]))] + [widgets.HTML("Rendering...") for _ in range(len(list_) - 1)]
     out = style(children=children)
     if labels is not None:
-        [out.set_title(i, label) for i, label in enumerate(labels)]
+        [out.set_title(i, str(label)) for i, label in enumerate(labels)]
```
CodyCBakerPhD, line 185: @bendichter This fixed it

I guess one of the labels in the test cases was actually an integer, and at some point they stopped magically casting it as a string during `set_title`

We should *probably* do it elsewhere we set titles as well just to be sure, but that can be separate PR

CodyCBakerPhD:
@bendichter This should be good to go and do a fresh release with - the drop in coverage seems to be because the reports on the main base aren't up to date (likely because this CI doesn't run on a CRON schedule) - that's something I can add in a follow-up

bendichter (approved):

PR #283, opened 2023-05-09
Fix fancy indexing for electrical series with Zarr backend

alejoe91:
Turns out fancy indexing is not supported for `zarr.Dataset` objects and this was causing an issue in displaying timeseries correctly.

This small PR fixes it by changing the behavior of the `timeseries.py` and:
- use direct slicing when possible (consecutive indices)
- slice first in time and then in channels otherwise

Not that for NWBwidgets in all cases we can slice directly, because of the channel slider.


bendichter (approved):

bendichter:
@alejoe91 it looks like Zarr does actually support fancy indexing, though the syntax is different from numpy

https://zarr.readthedocs.io/en/stable/tutorial.html#advanced-indexing

bendichter:
Yes that would be an immediate fix. But this issue is going to come up for anything downstream of pynwb so maybe it makes sense to fix on the pynwb level. Maybe we could wrap all arrays in xarray?

nwbwidgets/timeseries.py
```diff
@@ -423,10 +423,15 @@ def _prep_timeseries(time_series: TimeSeries, time_window=None, order=None):
...
 
     if len(time_series.data.shape) > 1:
-        mini_data = time_series.data[t_ind_start:t_ind_stop, unique_sorted_order][:, inverse_sort]
+        unique_sorted_order, inverse_sort = np.unique(order, return_inverse=True)
+        # fancy indexing is not supported in zarr, so we use slice when possible
+        if np.all(np.diff(unique_sorted_order) == 1):
+            unique_sorted_order = slice(unique_sorted_order[0], unique_sorted_order[-1] + 1)
+            mini_data = time_series.data[t_ind_start:t_ind_stop, unique_sorted_order][:, inverse_sort]
+        else:
+            mini_data = np.array(time_series.data[t_ind_start:t_ind_stop])[:, unique_sorted_order][:, inverse_sort]
```
bendichter, line 434: can we add a check here and have this work differently only for Zarr dataset objects? I'd prefer to use the simultaneous indexing approach for h5py datasets where we can so we don't load data into memory when we don't need to. I also think this could and probably should be refactored into a data utility that can be used in other places

bendichter (changes requested):
@alejoe91 see comments

bendichter:
@alejoe91 can you also please add checks that this data indexing utility does indeed solve the problem for Zarr dataset objects?

PR #288, opened 2023-06-10
Update deprecated numpy types

bjhardcastle:
* exception raised as of numpy 1.24.0 for using `np.float`:
https://numpy.org/doc/stable/release/1.24.0-notes.html
* `np.integer` still okay to check against abc for all numpy ints
* also searched master branch for any use of: `np.object`, `np.bool`, `np.float`, `np.complex`, `np.str`, `np.int`
* found a use of dtype check against `np.float64` in `nwbwidgets/misc.py`, which is still okay, but updated anyway to include built-in and all numpy floating point types

### DIFF
--- a/nwbwidgets/brains.py
+++ b/nwbwidgets/brains.py
@@ -101,7 +101,7 @@ class HumanElectrodesPlotlyWidget(widgets.VBox):
             colors = group_inv
             show_leg = True
             show_scale = False
-        elif isinstance(electrodes[color_by][0], (np.ndarray, np.float)):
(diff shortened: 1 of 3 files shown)


bendichter (approved):

bjhardcastle:
Apologies, I missed one. 

Or rather, the search function on the `github.**dev**` view of a repository is either not exhaustive, or slow enough it can look finished when it's not!

I re-checked for all of the types - hopefully that is all of them now.

CodyCBakerPhD (approved):

bendichter:
@bjhardcastle thanks for the contribution!

PR #291, opened 2023-06-15
Control cache

bendichter:


luiztauffer (approved):
tested it and it looks good

PR #299, opened 2023-06-27
Docker/Panel fixes and improvements

luiztauffer:
- [x] fix requirements for Dockerfile
- [x] improve docker/readme
- [x] github action to build and publish Image on new releases
- [x] local mount option, depends on #300


bendichter:
lgtm! @CodyCBakerPhD could you take a look as well since you know more about this type of devops stuff?

CodyCBakerPhD:
Yes, I've been keeping my eye on it and will make some comments later as I notice a couple things while adding similar stuff to https://github.com/catalystneuro/neuroconv/pull/383

luiztauffer:
@CodyCBakerPhD any specific things you see that should be blocking this merge? I wanted to get the dockerized mode updated and start releasing the images alongside the pypi releases, it will make it very easy to run Panel remotely always with the latest updates

.github/workflows/auto-publish.yml
```diff
@@ -29,3 +29,28 @@ jobs:
...
+        uses: actions/checkout@v2
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
```
CodyCBakerPhD, line 48: Please use a `secrets.GITHUB_USERNAME`, the `github.actor` is whoever originally pushes the workflow, which may not be the same user the `GITHUB_TOKEN` credentials match

.github/workflows/auto-publish.yml
```diff
@@ -1,14 +1,14 @@
...
-name: Upload Package to PyPI
+name: Upload Package to PyPI and Publish NWBPanel Docker Image
 
 on:
   release:
     types: [published]
 
 jobs:
-  deploy:
+  publish-pypi:
```
CodyCBakerPhD, line 11: Highly recommend setting this to `push` for a commit or 2 on this PR to test/ensure that it works before merging it to main, otherwise it's going to be a hassle to debug

CodyCBakerPhD: (and then remove `push` before merging, of course)

.github/workflows/auto-publish.yml
```diff
@@ -29,3 +29,28 @@ jobs:
...
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
```
CodyCBakerPhD, line 49: Recommend using a static token that matches the specified username to keep them in sync - note that it will need package permissions as well

docs/source/contents/running_with_docker.md
```diff
@@ -0,0 +1,33 @@
+# Running with Docker
+
+## Use the pre-built image
+You can easily run NWB widgets Panel using Docker.
+
+```bash
+$ docker pull ghcr.io/NeurodataWithoutBorders/nwbwidgets-panel:latest
+$ docker run -p 8866:8866 nwbwidgets-panel
```
CodyCBakerPhD, line 8: I'd love to test this - can you either get this to trigger from this branch - or manually push your build [[instructions](https://github.com/catalystneuro/neuroconv/pull/383/files#diff-fcc22aa372714b7faa21a54039c3a7fbe1abd106c1f15f9f8e6cf85588dfe592R28-R47)]

docker/Dockerfile
```diff
@@ -1,16 +1,21 @@
 FROM python:3.10
```
CodyCBakerPhD, line 1:
```suggestion
FROM python:3.10
LABEL org.opencontainers.image.source=https://github.com/NeurodataWithoutBorders/nwbwidgets
```

The GHCR likes having these set

docker/Dockerfile_publish
```diff
@@ -0,0 +1,17 @@
+FROM python:3.10
```
CodyCBakerPhD, line 1:
```suggestion
FROM python:3.10
LABEL org.opencontainers.image.source=https://github.com/NeurodataWithoutBorders/nwbwidgets
```

docker/Dockerfile
luiztauffer: @CodyCBakerPhD can you send me some reference for this? I'm interested in why that is the case, might use it in other places as well if it's important

.github/workflows/auto-publish.yml
luiztauffer: I agree, will do that

docker/Dockerfile
CodyCBakerPhD: https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#labelling-container-images

CodyCBakerPhD: description and licenses also not a terrible idea I suppose but I haven't messed with that yet

luiztauffer: but setting these helped solve any problem, in specific? I still can't see a clear advantage of doing that
usually we can tag the same image with multiple targets, e.g. if we want to push to ECR, Dockerhub, Azure CR, etc... and in some cases that might be interesting to do. Why would we want to fix that label to ghcr?

CodyCBakerPhD: image

When I did it to my images, everything worked as expected. I believe I recall seeing errors otherwise

bendichter (dismissed):

nwbwidgets/version.py
```diff
@@ -1 +1 @@
-version = "0.10.2"
+version = "0.11.1"
```
luiztauffer, line 1:
```suggestion
version = "0.11.3"
```

.github/workflows/auto-publish.yml
```diff
@@ -1,14 +1,17 @@
 # This workflow will upload a Python Package using Twine when a release is created
 # For more information see: https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries
 
-name: Upload Package to PyPI
+name: Upload Package to PyPI and Publish NWBPanel Docker Image
 
 on:
+  # push:
+  #   branches:
+  #     - panel-fixes-and-improvements
```
luiztauffer, line 9:
```suggestion
  push:
    branches:
      - panel-fixes-and-improvements
```

luiztauffer, line 9:
```suggestion
  push:
    branches:
      - panel-fixes-and-improvements
```

bendichter (approved):

PR #316, opened 2023-10-10
enable/disable caching

luiztauffer:
Example:

image

@bendichter

### DIFF
--- a/nwbwidgets/panel.py
+++ b/nwbwidgets/panel.py
@@ -111,6 +111,19 @@ class Panel(widgets.VBox):
         elif args["new"] == "Local file":
             self.create_components_local_file_source()
 
+    def create_cache_row(self):
+        """Create cache row"""
+        if self.enable_cache:
+            self.cache_checkbox = widgets.Checkbox(description="cache")
+            self.cache_checkbox.observe(self.toggle_cache)
+            self.cache_path_text = widgets.Text("nwb-cache")
+            self.cache_path_text.layout.visibility = "hidden"
+            self.cache_row = widgets.HBox([self.cache_checkbox, self.cache_path_text])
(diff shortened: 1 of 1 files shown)


nwbwidgets/panel.py
```diff
@@ -127,18 +131,26 @@ def create_components_dandi_source(self, args=None):
...
+            self.cache_path_text = None
+            self.cache_row = widgets.HBox([])
+
+        self.source_dandi_file_button = widgets.Button(icon="check", description="Load file")
         self.source_dandi_vbox = widgets.VBox(
             children=[
                 self.source_dandi_id,
                 self.source_dandi_file_dropdown,
-                widgets.HBox([self.cache_checkbox, self.cache_path_text]),
+                self.cache_row,
```
bendichter, line 153: I'd prefer to just not include the cache row here, instead of making it an empty HBox.

nwbwidgets/panel.py
```diff
@@ -172,22 +182,30 @@ def create_components_s3_source(self):
...
-        self.cache_path_text.layout.visibility = "hidden"
+
+        if self.enable_cache:
+            self.cache_checkbox = widgets.Checkbox(
+                description="cache",
+            )
+            self.cache_checkbox.observe(self.toggle_cache)
+            self.cache_path_text = widgets.Text("nwb-cache")
+            self.cache_path_text.layout.visibility = "hidden"
+            self.cache_row = widgets.HBox([self.cache_checkbox, self.cache_path_text])
```
bendichter, line 193: This code is repeated. Could we factor it out into a private method?

luiztauffer:
@bendichter done

bendichter (approved):

